PR #68, opened 2023-04-23
fix: typeahead now making correct queries and show correct counts

ChaoticTempest:
This fixes a couple typeahead issues and does some new stuff:
- No longer doing an omni search since that doesn't really give you per the facet search that we want, so instead we have a single query per facet. This does scale super well with more facets, but that's fine for now since we only have like 4. With NLP we can improve this later too.
- display counts are correct now per search group/facet
- No results showing properly now too


doriancrutcher (approved):
These fixes look good! Tested out the changes and they look good!

PR #102, opened 2023-05-09
Edit Posts

gagdiez:
This PR adds the ability to edit posts (and only posts, no comments), the idea is that we can discuss if we like this approach.

### DIFF
--- a/src/Posts/Edit.jsx
+++ b/src/Posts/Edit.jsx
@@ -1,0 +1,411 @@
+if (!context.accountId) {
+    return <></>;
+  }
+  
+  State.init({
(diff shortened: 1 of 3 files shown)


gagdiez:
Do we have any safe way to test this? So far I only tested it in the vscode extension.

calebjacob:
@gagdiez You can set up the `bos-loader` to run the viewer locally with these component changes. Check out the README.

gagdiez:
thanks @calebjacob, tested it there and everything seams to work as expected! Now people will be able to correct typos

calebjacob (approved):

PR #106, opened 2023-05-17
Generate type declarations

shelegdmitriy:
Closes: https://pagodaplatform.atlassian.net/browse/DEC-1127

Generated types for:
- [x] Social methods;
- [x] State methods;
- [x] Near;
- [x] Storage;
- [x] clipboard;
- [x] context (context.accountId, context.networkId);

### DIFF
--- a/vm.d.ts
+++ b/vm.d.ts
@@ -1,0 +1,126 @@
+/**
+ * VM provides a convenient API to update the state of the component.
+ * More info {@link https://docs.near.org/bos/api/state#state-apis}
+ */
+declare namespace State {
+  /**
+   * `State.init` takes an object as an argument and initializes the state of the component with this object. It'll be no-op if the state is already initialized.
+   * More info {@link https://docs.near.org/bos/api/state#stateinit}
+   * @param state an initial state object for the component.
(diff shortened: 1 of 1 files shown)


calebjacob:
@shelegdmitriy Excellent start! Great work. There's one catch to keep in mind for `styled-components` when you do implement the types:

Due to BOS not allowing typical imports, we had to get creative with `styled-components` + `Radix`. So we came up with some additional syntax that is valid (passing a string that matches the name of a Radix component):

```
const SwitchThumb = styled("Switch.Thumb")`
  display: block;
  width: 21px;
  height: 21px;
  background-color: pink;
  border-radius: 9999px;
  box-shadow: 0 2px 2px var(--blackA7);
  transition: transform 100ms;
  transform: translateX(2px);
  will-change: transform;

  &[data-state='checked'] {
    transform: translateX(19px);
  }
`;
```

So the first step would be bringing in `styled-components` base types, then extending it to also allow passing a string that matches a Radix component.

mpeterdev (approved):

calebjacob (approved):

PR #107, opened 2023-05-19
DO NOT MERGE: Testnet deployment

milenakrawczyk:


.github/workflows/deploy-stage.yml
mpeterdev, line 1: lets standardize our workflow naming to make it more clear the environments we are deploying to. We should have 

```
deploy-dev-testnet (develop branch to discom-dev.testnet)
deploy-dev-mainnet (develop branch to ???)

deploy-prod-testnet (main branch to discom.testnet)
deploy-prod-mainnet(main branch to near)
```

Does that seem accurate?

mpeterdev: there is one potential downside of having non-prod components deployed to mainnet, if users come across them and interact with them it is possible they will write data to their SocialDB key for that component w/ a schema which is not final. This can lead to some migration headaches, but it is also a general concern anyone may have to deal with. I am leaning towards not attempting to solve for that at this time and instead spinning up a new effort to consider data integrity/migration best practices

mpeterdev: cc @charleslavon @AustinBaggio

milenakrawczyk: @mpeterdev I think that we shouldn't have `deploy-dev-mainnet` workflow. Anything that is posted there will be also visible on the `deploy-prod-mainnet`.

charleslavon: I agree with @milenakrawczyk 👍🏾

elliotBraem:
Hey, I want to improve [DevHub CI/CD](https://github.com/near/neardevhub-widgets/issues/304) workflow and have been taking a lot of inspiration from your guys' flow, it's well done!

Question though, I see that there is a deploy-dev-testnet and deploy-prod-testnet. When I read the comments above, it makes sense that deploy-prod-testnet should be a testnet deploy of main branch. However, it is triggered on changes to the "dev" branch -- seems to me like deploy-dev and deploy-prod are always the same code. Is this right?

milenakrawczyk:
Hey @elliotBraem! The difference between the deploy-dev and deploy-prod is the replacements file that is used: see [dev-testnet](https://github.com/near/near-discovery-components/blob/develop/.github/workflows/deploy-dev-testnet.yml#L21) and [prod-testnet](https://github.com/near/near-discovery-components/blob/develop/.github/workflows/deploy-prod-testnet.yml#L21). This means that the widgets are deployed to different accounts (_discom-dev.testnet_ and _discom.testnet_ respectively).

elliotBraem:
Thanks for responding @milenakrawczyk! 

Yeah -- I see that, but I'm kinda confused then what's the point? The same code (develop branch) is being deployed to testnet twice, just under two different account ids. 

I think there should be a fix where prod-testnet workflow tracks changes to the **main** branch; so discom-dev is a testnet deploy of "develop" and discom is a testnet deploy of "main". See PR: #411

milenakrawczyk:
You are right, @elliotBraem. Thanks for pointing that out and for the PR!

PR #108, opened 2023-05-25
Different button for mobile

marcinbodnar:


src/NearOrg/HomePage.jsx
```diff
@@ -434,15 +452,26 @@ return (
...
+          <Widget
+            src="near/widget/DIG.Button"
+            props={{
+              href: "/signup",
+              label: "Create Account",
+              variant: "affirmative",
+              size: "large",
+            }}
+          />
+        </TryItNow>
```
shelegdmitriy, line 474: Overall it's fine to go this way but I believe we can do much simpler: Let's remove `TryItNow` and use this [classes](https://getbootstrap.com/docs/5.3/utilities/display/#hiding-elements) to handle visibility for this buttons. I mean something like that:
```suggestion
          <Widget
            src="near/widget/DIG.Button"
            props={{
              href: "/onboarding",
              label: "Try It Now",
              variant: "affirmative",
              size: "large",
              className: "d-none d-md-inline-flex"
            }}
          />
          <Widget
            src="near/widget/DIG.Button"
            props={{
              href: "/signup",
              label: "Create Account",
              variant: "affirmative",
              size: "large",
              className: "d-inline-flex d-md-none"
            }}
          />
```
and you'll get the same result without custom breakpoints

shelegdmitriy (changes requested):

src/NearOrg/HomePage.jsx
marcinbodnar: @shelegdmitriy Thanks for the proposition. But, I don't think it's simpler, also bootstrap has a breakpoint at 960px, and we have 900px.

Also closing this type of situation (when you display one element or another) in a parent element that is controlling what is displayed is a good approach.

shelegdmitriy (approved):

PR #113, opened 2023-06-05
Share Button

calebjacob:
Adds a share button to each post and comment. The copy URL action/button was removed from the dropdown since it already exists at the top level (next to the share button).


mpeterdev (approved):
LGTM

PR #115, opened 2023-06-08
Fix error on Onboarding.ComponentCollection

shelegdmitriy:
Closes: https://pagodaplatform.atlassian.net/browse/DEC-1172

This error occurs only for. account with zero widgets.

### DIFF
--- a/src/Onboarding/ComponentCollection.jsx
+++ b/src/Onboarding/ComponentCollection.jsx
@@ -1,6 +1,7 @@
 const accountId = props.accountId || context.accountId || "near";
 
 const profile = Social.getr(`${accountId}/profile`);
+const widgets = Social.getr(`${accountId}/widget`) ?? {};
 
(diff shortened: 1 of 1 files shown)


src/Onboarding/ComponentCollection.jsx
```diff
@@ -12,14 +13,6 @@ const allWidgetsHistoryChangesBlocks = Social.keys(
 
 if (allWidgetsHistoryChangesBlocks === null) return "Loading...";
 
-const widget = allWidgetsHistoryChangesBlocks[accountId].widget;
-
-const totalCommits = Object.keys(widget)
-  .map((key) => widget[key])
-  .flat();
-
-const widgets = Social.getr(`${accountId}/widget`) ?? {};
```
shelegdmitriy, line 21: Actually we haven't use this variables so it's just cleanup

PR #116, opened 2023-06-14
adding NEARWEEK components

AngelBlock:


charleslavon (changes requested):
Please move these into a new folder called `NearWeek`, and rename each file, dropping the NearWeek prefix.

AngelBlock:
> Please move these components into a new folder called `NearWeek`, and rename each component, dropping the NearWeek prefix.

This is done as requested. Please let us know if you need anything else

charleslavon (approved):

src/NEARWEEK/News.jsx
```diff
@@ -0,0 +1,267 @@
...
+      <Wrapper>
+        <H2>News</H2>
+        <Content>
+          {news.map((item, index) => (
+            <Card index={index}>
+              <div class="d-flex">
+                <CardImage width="78" height="78" src={item.thumbnail} alt="" />
+                <CardContent>
+                  <CardTitle>
+                    <a href={article.url} target="_blank">
```
charleslavon, line 236: @AngelBlock looks like these URLs are not coming through as expected. I'm seeing them open a tab to `about:blank`
Screen Shot 2023-06-20 at 4 40 54 PM

AngelBlock: Apologies, this is honest mistake, with 4 sources of code, it's hard to maintain. The code was updated to the last known version and PR created #120

charleslavon:
@AngelBlock Not a problem at all!  Actually you bring up a good point that's related to a conversation I just had with Gautham We were considering What minimal process should we follow to allow community contributed components to be added into the near.org gateway, and when should components be added into the root near account?

For the goal of getting these new NearWeek components displayed in the near.org gateway, it's not necessary that they be added into the root near account (which is what this PR would accomplish). Rather, they can be added directly into the near.org's ActivityFeed/ExploreWidget component. This would avoid having your components existing in multiple places, and make it clear via the gateway URL and activity feed component, that community contributions are welcome and highlighted.

It will so simply our guidance on when components should be deployed onto the root Near account - when they are a design system or default component of the near.or gateway. 

With that in mind, I'm going to revert this PR.  Please follow-up with a new PR which updates [ExploreWidgets](https://github.com/near/near-discovery-components/blob/develop/src/ExploreWidgets.jsx#L11) to reference the appropriate NearWeek news component.

AngelBlock:
> @AngelBlock Not a problem at all! Actually you bring up a good point that's related to a conversation I just had with Gautham We were considering What minimal process should we follow to allow community contributed components to be added into the near.org gateway, and when should components be added into the root near account?
> 
> For the goal of getting these new NearWeek components displayed in the near.org gateway, it's not necessary that they be added into the root near account (which is what this PR would accomplish). Rather, they can be added directly into the near.org's ActivityFeed/ExploreWidget component. This would avoid having your components existing in multiple places, and make it clear via the gateway URL and activity feed component, that community contributions are welcome and highlighted.
> 
> It will so simply our guidance on when components should be deployed onto the root Near account - when they are a design system or default component of the near.or gateway.
> 
> With that in mind, I'm going to revert this PR. Please follow-up with a new PR which updates [ExploreWidgets](https://github.com/near/near-discovery-components/blob/develop/src/ExploreWidgets.jsx#L11) to reference the appropriate NearWeek news component.

Hi, that's what I initially thought would happen. Similar to this page https://near.org/denysk.near/widget/ActivityPage, you would simply include NEARWEEK components on the home page. This would undoubtedly simplify many things and assist with component maintenance. If this can be achieved with a PR (while maintaining the quality of the homepage, of course), it's superb.

I'll review the new process and revert if I have any questions.

PR #120, opened 2023-06-21
NEARWEEK News.jsx link fix

AngelBlock:


src/NEARWEEK/News.jsx
```diff
@@ -221,6 +226,8 @@ function dateToDays(date) {
   return timeAgo(Date.now() - d.getTime());
 }
 
+console.log(news);
```
milenakrawczyk, line 229: Is it intentional?

milenakrawczyk (commented):
Thank you for the contribution! I left one comment.

src/NEARWEEK/News.jsx
AngelBlock: no, its not necessary, removed here https://github.com/near/near-discovery-components/pull/120/commits/013dad164b1385cdee72733e5adb36cf8b7fc5d0 thanks for checking

PR #124, opened 2023-06-23
Update ExploreWidgets.jsx

AngelBlock:
Default to nearweekapp.near/widget/nearweek-newsletter

### DIFF
--- a/replacements.dev.json
+++ b/replacements.dev.json
@@ -1,7 +1,8 @@
 {
     "REPL_ACCOUNT": "discom-dev.testnet",
-    "REPL_NEAR_URL": "test.near.org",
     "REPL_MOB": "eugenethedream",
     "REPL_MOB_2": "one.testnet",
     "REPL_MODERATOR": "bosmod.testnet"
+    "REPL_NEAR_URL": "test.near.org",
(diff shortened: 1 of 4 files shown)


AngelBlock:
@milenakrawczyk I'm not sure if this is the best way to handle this, created PR anyway, but lets discuss further

charleslavon:
@AngelBlock take a look at my last commit. This is to support deploy components across distinct mainnet/testnet contracts as described in this [recently updated doc](https://github.com/near/near-discovery-components#contributing).   If you're currently using a specific testnet account to test nearweekapp components, please add it [here](https://github.com/near/near-discovery-components/pull/124/files#diff-24a201ec7d9862267be6b3c9ed01660d7e73e0e30e0674aae9f31d0c9a028209R7) and [here](https://github.com/near/near-discovery-components/pull/124/files#diff-1655b11169a2ad30a2b15f9bdd3850d9955c0179d2baf3fad982c1b7214559edR7), if not we'll deploy onto our discom-dev.testnet.

AngelBlock:
@charleslavon I will review and revert earlier next week, thanks for sharing.

charleslavon:
Merging so we can preview this on test.near.org

PR #126, opened 2023-06-28
DO NOT MERGE: Add chaotictempest to replacements

milenakrawczyk:


milenakrawczyk:
@shelegdmitriy I am closing this PR since the chaotictempest components were moved to our repo in this PR: https://github.com/near/near-discovery-components/pull/134 so we don't need to add separate replacements.

PR #131, opened 2023-06-29
Change desktop minimal width

shelegdmitriy:
It's not so clear which element should change depending on mobile or desktop. I tested it locally and it seems to be OK but still need some eyes on it.

Closes: https://pagodaplatform.atlassian.net/browse/DEC-1318
Relates to: https://github.com/near/near-discovery/pull/280


calebjacob (approved):

charleslavon:
testing LGTM!

PR #152, opened 2023-07-26
DEC-1349: Add the ability to complain about user profiles

shelegdmitriy:
Closes: https://pagodaplatform.atlassian.net/browse/DEC-1349

This is how `FlagButton` looks like on profile page:

Знімок екрана 2023-07-28 о 15 08 44

and on tooltip:

Знімок екрана 2023-07-28 о 15 08 57

From now we using the same `AccountProfileOverlay` component everywhere.

### DIFF
--- a/src/AccountProfileInline.jsx
+++ b/src/AccountProfileInline.jsx
@@ -31,6 +31,7 @@ const Wrapper = styled.a`
 `;
 
 const Text = styled.span`
+  display: inline-flex;
   margin: 0;
   font-size: 14px;
   line-height: 20px;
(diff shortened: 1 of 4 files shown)


src/AccountProfileInline.jsx
```diff
@@ -31,6 +31,7 @@ const Wrapper = styled.a`
 `;
 
 const Text = styled.span`
+  display: inline-flex;
```
calebjacob, line 34: This is going to break text truncation on overflow for really long account names. I had to remove `display: inline-flex;` here on a previous commit to fix it. What caused the need to add it back?

calebjacob (changes requested):

src/AccountProfileInline.jsx
shelegdmitriy: This change was done by mistake after comparing with your's copy of `AccountProfileInline`

calebjacob (approved):

PR #159, opened 2023-08-03
DEC-1352: Add confirmation modal on flagging account for moderation

shelegdmitriy:
Closes:
- https://pagodaplatform.atlassian.net/browse/DEC-1352;
- #315


calebjacob:
@shelegdmitriy Let's make use of the Radix Dialog primitive for the modal: https://www.radix-ui.com/docs/primitives/components/dialog

shelegdmitriy:
@calebjacob indeed! I left this is this state to focus on main functional part. As soon as we'll get a design for DIG.Modal I'll fix this

charleslavon:
hey @shelegdmitriy To make review/merging easier,  use the ability to save/create a PR as `Draft` as oppose to writing Don't Merge in the description.

src/Flagged/Modal.jsx
```diff
@@ -0,0 +1,97 @@
...
+  const flaggedAccountsListSet = new Set(flaggedAccountsList);
+  const filterFlaggedAccounts = [
+    ...flaggedAccountsListSet.add(reportedAccountId),
+  ];
+
+  Social.set(
+    {
+      index: {
+        moderate: {
+          hide: {
```
gabehamilton, line 49: Can you switch to the new structure described [here](https://github.com/near/near-discovery/discussions/386)?

```suggestion
         [filterFlaggedAccounts]: ""
```

shelegdmitriy: And it will mean that we want ti hide all the posts without reporting, right @gabehamilton ?

gabehamilton: Yes, that's correct.

src/AccountProfile.jsx
```diff
@@ -1,6 +1,7 @@
 const accountId = props.accountId || context.accountId;
 const profile = props.profile || Social.get(`${accountId}/profile/**`, "final");
 const profileUrl = `#/${REPL_ACCOUNT}/widget/ProfilePage?accountId=${accountId}`;
+const activityFeed = props.activityFeed;
```
gabehamilton, line 4: maybe this prop should be named something like `showActionButtons`?

shelegdmitriy: This prop defines the place where FlagButton is clicked. We need to be sure that flag account button is clicked on ActivityFeed

gabehamilton: Right. I was thinking we may show the AccountProfile lots of places. With some we'll want to show those buttons (like on the ActivityFeed), others we won't.

shelegdmitriy: Make sense

gabehamilton:
I think the Confirm and Cancel buttons need more functionality. Both lead to the save data modal and they need to close the 'Do you want to hide all posts of this account?' modal so you can interact with the save data modal.

gabehamilton:
Looks really good. Normally I'd say it's ready to merge. However, I just re-opened the data discussion https://github.com/near/near-discovery/discussions/386
Let's wait to merge this PR until we can resolve that discussion and store the hide account data in the final format.

shelegdmitriy:
After a small discussion with @gabehamilton I decided to add a couple if things here:

1. "Not Now" button or "X" button to simply close this Dialog without reporting.
2. Add blur effect on background. This change makes this Dialog to looks exactly like NotificationAlert (I've added blur effect by default but this can be controlled by developer).

All this things could be added in just 1 line of code here after #357 will be merged.

cc @charleslavon

src/Flagged/Modal.jsx
```diff
@@ -0,0 +1,71 @@
...
+  const flaggedAccountsList = getFlaggedAccountsList
+    ? JSON.parse(getFlaggedAccountsList)
+    : [];
+  const flaggedAccountsListSet = new Set(flaggedAccountsList);
+  const filterFlaggedAccounts = [
+    ...flaggedAccountsListSet.add(reportedAccountId),
+  ];
+
+  socialSet({
+    index: {
```
gabehamilton, line 49: This data won't go in the index. You can take out this line (and the matching brace.

src/Flagged/Modal.jsx
```diff
@@ -0,0 +1,71 @@
...
+  ...forwardedProps
+} = props;
+
+const profileUrl = `#/${REPL_ACCOUNT}/widget/ProfilePage?accountId=${reportedAccountId}`;
+
+let getFlaggedAccountsList;
+
+// copied from Moderate.jsx so maybe it's wrong
+if (forwardedProps.open) {
+  getFlaggedAccountsList = Social.get(
```
gabehamilton, line 14: You don't need the existing accounts (or posts). Each user's self-moderation data is going into the moderate object with the accounts as keys. The `set` method merges keys into that object when it stores it in the contract.

src/Flagged/Modal.jsx
```diff
@@ -0,0 +1,71 @@
...
+    : [];
+  const flaggedAccountsListSet = new Set(flaggedAccountsList);
+  const filterFlaggedAccounts = [
+    ...flaggedAccountsListSet.add(reportedAccountId),
+  ];
+
+  socialSet({
+    index: {
+      moderate: {
+        [filterFlaggedAccounts]: "",
```
gabehamilton, line 51: Let's make this `report` for now. It can get more options when we do the reporting flow. In that flow `report` will be the value to store for a generic report (without choosing a reason).
```suggestion
        [filterFlaggedAccounts]: "report",
```

src/Flagged/Modal.jsx
shelegdmitriy: But why? We've been making a `Social.set()` all the time. @gabehamilton could you please provide more info here?

gabehamilton: Latest update. Let's use this format for compatibility with gateways that may want to use Social.index in their implementation.
```
    "index": {
      "moderation": `{\"key\":\"reported\",\"value\":{\"path\":\"${reportedAccountId}\", \"label\": \"report\"}}`
    },
    "moderation": {
      [reportedAccountId]: "report"
    },
```

gabehamilton (approved):

src/Flagged/Modal.jsx
```diff
@@ -0,0 +1,51 @@
...
+  });
+};
+
+const reportAccountWithPosts = () => {
+  socialSet({
+    index: {
+      moderate: {
+        [reportedAccountId]: "report",
+      },
+    },
```
shelegdmitriy, line 33:
```suggestion
    index: {
      moderation: {
        key: "reported",
        value: {
          path: reportedAccountId,
          label: "report"
        },
        [reportedAccountId]: "report",
      },
    },
    moderation: {
      [reportedAccountId]: "report",
    },
```


@gabehamilton according to your comment below is this what you meant?

gabehamilton: The index.moderation value needs to be a string for the index standard.
```suggestion
      index: {
          moderation: `{"key":"reported","value":{"path":"${reportedAccountId}", "label": "report"}}`
      },
      moderation: {
          [reportedAccountId]: "report",
      },

```

src/AccountProfile.jsx
roshaans: ll

src/AccountProfile.jsx
```diff
@@ -87,13 +89,31 @@ const AccountProfile = (
...
... 29 17.2837 28.0612 18.1475 27.2325C18.64 26.7612 19.1475 26.2738 19.5938 26.09C20.0025 25.92 20.68 25.91 21.3363 25.9C22.5563 25.8812 23.9387 25.8612 24.9 24.9C25.8612 23.9387 25.8812 22.5563 25.9 21.3363C25.91 20.68 25.92 20.0025 26.09 19.5938C26.2738 19.1488 26.7613 18.64 27.2325 18.1475C28.0613 17.285 29 16.305 29 15C29 13.695 28.0613 12.7162 27.2325 11.8525ZM20.7075 11.2925C20.8005 11.3854 20.8742 11.4957 20.9246 11.6171C20.9749 11.7385 21.0008 11.8686 21.0008 12C21.0008 12.1314 20.9749 12.2615 20.9246 12.3829C20.8742 12.5043 20.8005 12.6146 20.7075 12.7075L13.7075 19.7075C13.6146 19.8005 13.5043 19.8742 13.3829 19.9246C13.2615 19.9749 13.1314 20.0008 13 20.0008C12.8686 20.0008 12.7385 19.9749 12.6171 19.9246C12.4957 19.8742 12.3854 19.8005 12.2925 19.7075L9.2925 16.7075C9.10486 16.5199 8.99944 16.2654 8.99944 16C8.99944 15.7346 9.10486 15.4801 9.2925 15.2925C9.48014 15.1049 9.73464 14.9994 10 14.9994C10.2654 14.9994 10.5199 15.1049 10.7075 15.2925L13 17.5863L19.2925 11.2925C19.3854 11.1995 19.4957 11.1258 19.6171 11.0754C19.7385 11.0251 19.8686 10.9992 20 10.9992C20.1314 10.9992 20.2615 11.0251 20.3829 11.0754C20.5043 11.1258 20.6146 11.1995 20.7075 11.2925Z" fill="#202425"/>
```
roshaans, line 94: creating a thread with gh.nvim

src/Flagged/Modal.jsx
shelegdmitriy: Thanks @gabehamilton!

PR #161, opened 2023-08-03
DEC-1370 feat: port queryapi feed -> near-discovery-components

roshaans:
near.org feed powered with QueryAPI

### DIFF
--- a/src/Feed.jsx
+++ b/src/Feed.jsx
@@ -1,5 +1,122 @@
+const GRAPHQL_ENDPOINT =
+  props.GRAPHQL_ENDPOINT || "https://near-queryapi.api.pagoda.co";
+const LIMIT = 25;
+let accountsFollowing = props.accountsFollowing;
+const moderatorAccount = props?.moderatorAccount || "bosmod.near";
+
+State.init({
+  posts: [],
+  postsCountLeft: 0,
+  initLoadPosts: false,
+  initLoadPostsAll: false,
+});
+
+let filterUsersRaw = Social.get(
+  `${moderatorAccount}/moderate/users`,
+  "optimistic",
+  {
+    subscribe: true,
+  }
+);
+
+if (filterUsers === null) {
+  // haven't loaded filter list yet, return early
+  return "";
+}
+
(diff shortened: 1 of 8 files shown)


shelegdmitriy (changes requested):

src/Feed.jsx
```diff
@@ -1,29 +1,122 @@
...
+
+let filterUsersRaw = Social.get(
+  `${moderatorAccount}/moderate/users`,
+  "optimistic",
+  {
+    subscribe: true,
+  }
+);
+
+if (filterUsers === null) {
```
mpeterdev, line 22: this should be `filterUsersRaw`, it was an existing bug

src/Posts.jsx
```diff
@@ -1,32 +1,235 @@
...
 if (previousSelectedTab && previousSelectedTab !== state.selectedTab) {
   State.update({
     selectedTab: previousSelectedTab,
   });
 }
 
-let accounts = undefined;
-
-if (state.selectedTab === "following" && context.accountId) {
+// TODO port to QueryAPI
```
mpeterdev, line 23: this is intentionally left for later, right?

src/Posts.jsx
```diff
@@ -1,32 +1,235 @@
...
+  }
+);
+
+const selfFlaggedPosts = context.accountId
+  ? Social.index("flag", "main", {
+      accountId: context.accountId,
+    })
+  : [];
+
+if (filterUsers === null) {
```
mpeterdev, line 61: should also be `filterUsersRaw`

mpeterdev (changes requested):
code changes LGTM, only change requested is fixing the existing bug I pointed out while you're in here

still need to do more functional testing, currently seeing `VM is dead errors` when running with bos-loader

src/Posts.jsx
roshaans: Yes. We fetch the followers using the `Social` API as we do not yet index follower with QueryAPI yet.

mpeterdev (approved):

PR #165, opened 2023-08-04
fix: Math.max with an undefined argument resulted in NaN where an integer was expected.

charleslavon:
This either fixes the React Minify error completely, or is a step towards that direction.

once this is deployed into `develop` we can test the discussions component @ https://test.near.org/discom.testnet/widget/ComponentDetailsPage?src=discom.testnet/widget/ActivityPage&tab=discussion


shelegdmitriy (approved):
Let's try 🙂

PR #175, opened 2023-08-11
Notification component

marcinbodnar:
It's not the final version of the component, but it's the proper layout for further work.


src/NearOrg/Notifications/Notification.jsx
```diff
@@ -0,0 +1,99 @@
...
+`;
+
+return (
+  <>
+    <Notification>
+      <Icon>
+        <svg
+          width="20"
+          height="20"
+          viewBox="0 0 20 20"
```
shelegdmitriy, line 70: You can simply use this icon instead:

```suggestion
      <i class="ph ph-git-fork" />
```

I check it here https://phosphoricons.com/ and it looks exact the same

shelegdmitriy (changes requested):

charleslavon:
@marcinbodnar should this PR be closed in favor of #330?

PR #177, opened 2023-08-11
Notifications List component

marcinbodnar:
It's not the final version of the component, but it's the proper layout for further work.


charleslavon:
@marcinbodnar should this PR be closed in favor of #330?

PR #180, opened 2023-08-15
lock TimeAgo dependency to previous version

### LINKED ISSUE #179: Fix "time ago" string generation for Featured Components
The "time ago" value used to always show a relative timestamp i.e. `1m ago` `3w ago` but now it displays a relative timestamp only for short diffs and a date for longer diffs. We are always appending the string `ago` to the time value, which leads malformed output when we get a date

This screenshot shows both types in effect
Screenshot 2023-08-15 at 9 42 44 AM

Solution options:

1. Revert to always showing relative time and continue appending `ago` to every entry
2. Determine the breaking point where the time display switches to date and omit `ago` in those cases

Both makes sense, it's a matter of UX preference

## Relevant Code
`near/widget/ComponentCard`
```jsx
<CardTag>
  <i className="bi bi-clock"></i>{" "}
  <Widget
    src="mob.near/widget/TimeAgo"
    props={{
      blockHeight: props.blockHeight,
      keyPath: `${accountId}/widget/${widgetName}`,
    }}
  />{" "}
  ago
</CardTag>
```

mpeterdev:
closes #179

### DIFF
--- a/src/ComponentCard.jsx
+++ b/src/ComponentCard.jsx
@@ -152,7 +152,7 @@ return (
       <CardTag>
         <i className="bi bi-clock"></i>{" "}
         <Widget
-          src="${REPL_MOB_2}/widget/TimeAgo"
+          src="${REPL_MOB_2}/widget/TimeAgo@97556750"
           props={{
             blockHeight: props.blockHeight,
             keyPath: `${accountId}/widget/${widgetName}`,


mpeterdev:
hotfix rolled out which cherry picked commit so this should be merged with a merge commit instead of squashing

shelegdmitriy (approved):

PR #187, opened 2023-08-16
feat: Trending Users component

scopalaffairs:
# Trending Users Components

The widget TrendingUsers is intended for the logged in user experience, i.e. placed above the news feed.
It displays a list of selected user profiles ranked by metrics from data fetched from a GCS bucket.

The adaption of PeoplePage is intended for both logged in as well as logged out UX.

For performance reasons, results are paginated and loaded on demand by clicking the load more button.
In the future, more recommendation components are going to be developed, that's why I opened a new directory `Recommendations`.


scopalaffairs:
Converted to draft, because we are revising the UI layout

PR #189, opened 2023-08-17
Fix posts list on profile page

shelegdmitriy:
Closes: #188

### DIFF
--- a/src/ProfilePage.jsx
+++ b/src/ProfilePage.jsx
@@ -240,7 +240,7 @@ return (
             )}
 
             <Widget
-              src="${REPL_ACCOUNT}/widget/Posts.Feed"
+              src="${REPL_ACCOUNT}/widget/v1.Posts.Feed"
               props={{ accounts: [accountId] }}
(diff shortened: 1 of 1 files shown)


src/ProfilePage.jsx
```diff
@@ -240,7 +240,7 @@ return (
             )}
 
             <Widget
-              src="${REPL_ACCOUNT}/widget/Posts.Feed"
+              src="${REPL_ACCOUNT}/widget/v1.Posts.Feed"
```
shelegdmitriy, line 243: I wonder is there any other places this can be broken?

mpeterdev (approved):

mpeterdev:
good catch. This is the right fix for now, can you create a ticket for the backlog to upgrade the account posts feed to QueryApi?

shelegdmitriy:
> good catch. This is the right fix for now, can you create a ticket for the backlog to upgrade the account posts feed to QueryApi?

Yes sure!

PR #201, opened 2023-08-30
Fix styling of `v1.LikeButton`

shelegdmitriy:
I like how this button shows users who liked the post, but here are a few things we need to fix:
- enable this for testnet (`<Widget src="mob.near/widget/LikeButton.Faces" props={{ likesByUsers }} />` will work for mainnet and it seems there is no such copy for testnet);
- improve styling of overlay so it will have the same styling as other components on near.org

Знімок екрана 2023-08-30 о 13 59 20

I think we can move on with this PR but let's create tasks for improvement.

cc @charleslavon @roshaans


shelegdmitriy:
I'll close this one in favour of #200

PR #205, opened 2023-09-05
Various DIG Fixes

calebjacob:
Fixed various issues with DIG components that I came across while doing further testing. Also updated some of the docs for more consistent formatting. These changes are in support of this ticket: https://pagodaplatform.atlassian.net/browse/DEC-1443


src/DIG/Tabs.jsx
```diff
@@ -245,8 +236,9 @@ return (
         </Tab>
       ))}
     </TabList>
-    <TabContent value={items[state.activeTab - 1].value}>
-      {items[state.activeTab - 1].content}
-    </TabContent>
+
+    {items.map((item) => (
+      <TabContent value={item.value}>{item.content}</TabContent>
+    ))}
```
shelegdmitriy, line 242: The reason why I didn't do it is performance. For example if every tab has to load some data? Using my version that data will load on choosing tab you need but using code you presented it will load all data for all the tabs and they will be hidden.

If you feel it's fine to go this way - I'm good with this

shelegdmitriy (approved):

src/DIG/Tabs.jsx
calebjacob: I ran into some edge cases with the previous implementation - so that's what lead to the change. This change also matches the pattern that Radix recommends. I believe Radix takes care of not rendering inactive tabs, so there shouldn't be too much of a performance change. 👍

PR #321, opened 2023-09-12
feat: Filter out accounts and posts that have been self moderated

gabehamilton:
Feed side of UI implemented in https://github.com/near/near-discovery-components/pull/159
See discussion https://github.com/near/near-discovery/discussions/386

### DIFF
--- a/src/Posts.jsx
+++ b/src/Posts.jsx
@@ -97,7 +97,7 @@ function matchesModeration(moderated, item) {
     return true;
   }
   // match posts
-  return accountFound[item.block_height] !== "undefined";
+  return typeof accountFound[item.block_height] !== "undefined";
 }
 
 const shouldFilter = (item) => {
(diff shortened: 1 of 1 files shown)


src/Posts.jsx
```diff
@@ -64,25 +64,52 @@ const selfFlaggedPosts = context.accountId
...
+function matchesModeration(moderated, item) {
+  let accountFound = moderated[accountId];
+  if (typeof accountFound === "undefined") {
+    return false;
+  }
+  if (typeof accountFound === "string") {
+    return true;
+  }
+  // match posts
+  return accountFound[item.block_height] !== "undefined";
```
gabehamilton, line 100:
```suggestion
  return typeof accountFound[item.block_height] !== "undefined";
```

shelegdmitriy (approved):

PR #330, opened 2023-09-15
Notifications components

marcinbodnar:


src/NearOrg/Notifications/NotificationsList.jsx
```diff
@@ -0,0 +1,274 @@
...
+  <div className="loader" key={"loader"}>
+    <span
+      className="spinner-grow spinner-grow-sm me-1"
+      role="status"
+      aria-hidden="true"
+    />
+    Loading ...
+  </div>
+);
+
```
charleslavon, line 230: @marcinbodnar can you please explain a bit why we need to query and build a feed in a similar way as `IndexFeed` and how does this fit into the push notification data stream?

PR #331, opened 2023-09-17
encourage the creation of custom profile components

charleslavon:
This PR updates ProfilePage to redirect to an accountId's custom profile page if it exists.

### DIFF
--- a/src/ProfilePage.jsx
+++ b/src/ProfilePage.jsx
@@ -10,7 +10,8 @@ if (Social.getr(`${accountId}/widget/ProfilePage`)) {
       src={`${accountId}/widget/ProfilePage`}
       props={{
         accountId,
-      }}/>
+      }}
+    />
   );
 }
 State.init({
--- a/src/ProfilePage.jsx
+++ b/src/ProfilePage.jsx
@@ -1,11 +1,21 @@
-State.init({
-  selectedTab: props.tab || "overview",
-});
-
 const accountId = props.accountId ?? context.accountId;
 if (!accountId) {
   return "No account ID";
 }
+
+// if it exists, render the accountId's custom profile component
+if (Social.getr(`${accountId}/widget/ProfilePage`)) {
+  return (
+    <Widget
+      src={`${accountId}/widget/ProfilePage`}
+      props={{
+        accountId,
+      }}/>
+  );
+}
+State.init({
+  selectedTab: props.tab || "overview",
+});
 
 if (props.tab && props.tab !== state.selectedTab) {
   State.update({


src/ProfilePage.jsx
```diff
@@ -1,12 +1,22 @@
...
 
+// if it exists, render the accountId's custom profile component
+if(Social.getr(`${accountId}/widget/ProfilePage`)){
+  return (<Widget 
+            src={`${accountId}/widget/ProfilePage`}
+            props={{
+              accountId
+            }}
+          />)
+}
```
shelegdmitriy, line 14: @charleslavon could you please apply formatting here?

shelegdmitriy (changes requested):

gabehamilton (approved):

gabehamilton:
Was thinking about this more. Maybe this could use a wrapper (tabs?) such that the viewing user can switch between the custom profile component and the standard one? The standard one may serve as a source of truth about users.

shelegdmitriy (approved):

shelegdmitriy:
@gabehamilton sounds interesting. But let's move this to the discussions, @charleslavon WDYT?

thisisjoshford:
uh-mazing @charleslavon ! 💪 Thank you so much for putting some time in on your weekend off to hack on this! 🙏

@gabehamilton I like your idea for adding more options here (re: tabs/toggle) and also have some ideas as far as wanting to standardize _some_ things on the profile page such as contact info, contribution graph, profile image, sponsor/tip button etc. This way the most common things people want to find are highly discoverable and people don't have to play "Where's Waldo" if they wanna reach out to a dev or tip them.

I'm a big fan of being Aglie and shipping small incremental improvements rather than blocking progress here. IMO we should ship this as is (or make a minor tweak if needed) and then think more deeply about this with the design team (@heycorwin @sarahmills) 

cc @ravigautham

thisisjoshford (approved):
uh-mazing @charleslavon ! 💪 Thank you so much for putting some time in on your weekend off to hack on this! 🙏

@gabehamilton I like your idea for adding more options here (re: tabs/toggle) and also have some ideas as far as wanting to standardize _some_ things on the profile page such as contact info, contribution graph, profile image, sponsor/tip button etc. This way the most common things people want to find are highly discoverable and people don't have to play "Where's Waldo" if they wanna reach out to a dev or tip them.

I'm a big fan of being Aglie and shipping small incremental improvements rather than blocking progress here. IMO we should ship this as is (or make a minor tweak if needed) and then think more deeply about this with the design team (@heycorwin @sarahmills) 

cc @ravigautham

PR #346, opened 2023-09-21
Notifications - show on hover notifications list

marcinbodnar:


shelegdmitriy (approved):

marcinbodnar:
@shelegdmitriy This PR will enable the hover element that you prepared. There are two small issues with the button and hover container:
 - the hover container is not aligned properly with the bell icon
 - when you hover over account name on notifications, another small hover with account info should be displayed, but it's hiding behind the main container.
 
There is also one thing that would be good to add:
 - when user clicks on Settings icon or anything that is redirecting, the container should be close.

shelegdmitriy:
> @shelegdmitriy This PR will enable the hover element that you prepared. There are two small issues with the button and hover container:
> 
> * the hover container is not aligned properly with the bell icon
> * when you hover over account name on notifications, another small hover with account info should be displayed, but it's hiding behind the main container.
> 
> There is also one thing that would be good to add:
> 
> * when user clicks on Settings icon or anything that is redirecting, the container should be close.

Thanks for the feedback @marcinbodnar! If those changes are not blocking from testing this feature I could fix those tomorrow if you don't mind?

shelegdmitriy:
I've added this items here https://github.com/near/near-discovery-components/issues/348 so we won't lose it

cc @charleslavon @marcinbodnar

PR #354, opened 2023-09-22
Notification path updates to remove '#'

charleslavon:
To avoid the use of '#' in URI prefixes as they cause an unnecessary redirects


charleslavon:
@marcinbodnar could you help verify this? I see that we use '#' in some places but not all. From my testing this seems to cause a delay in the redirect.

PR #357, opened 2023-09-26
Add more control to DIG.Dialog

shelegdmitriy:


src/DIG/Dialog.jsx
```diff
@@ -150,32 +173,36 @@ return (
...
-          </Dialog.Close>
-        )}
-
-        {type === "alert" && (
-          <AlertDialog.Action asChild>
-            <div className="d-inline-block">{confirmButton}</div>
-          </AlertDialog.Action>
+      {content}
+      <ActionWrapper>
+        {actionButtons ?? (
```
shelegdmitriy, line 181: Sometimes two buttons is not enough. For example I want to add three buttons:
- Yes;
- No;
- Not sure.
So I wonder is my solution is good enough or I have to think about it more? I was thinking about such example:

```jsx
{actionButtons}
{type === "alert" ? (
  <>
    <AlertDialog.Cancel asChild>
      <div className="d-inline-block">{cancelButton}</div>
    </AlertDialog.Cancel>
    <AlertDialog.Action asChild>
        <div className="d-inline-block">{confirmButton}</div>
      </AlertDialog.Action>
    </>
  ) : (
  <Dialog.Close asChild>
    <div className="d-inline-block">{confirmButton}</div>
  </Dialog.Close>
)}
```
In this example we can leave default buttons and add our own. And using `actionStyled` property we can change it's order or so.

@calebjacob WDYT? Should I leave this code as commited or change according to comment?

calebjacob: @shelegdmitriy Good question. I think I prefer the way the code is committed. If the default buttons aren't right for my scenario, I'd expect to pass in an override for the entire `ActionWrapper` section. This allows me to have complete control over the buttons (there might be a case where I don't want the cancel or confirm button).

calebjacob (approved):

src/DIG/Dialog.jsx
shelegdmitriy: Agree. Thanks @calebjacob!

PR #365, opened 2023-09-27
Fix new notifications count

shelegdmitriy:
Closes: [#576](https://github.com/near/near-discovery/issues/576)

### DIFF
--- a/src/NearOrg/Notifications/NotificationButton.jsx
+++ b/src/NearOrg/Notifications/NotificationButton.jsx
@@ -15,6 +15,11 @@ State.init({
   open: false,
 });
 
+const handleDropdownOpen = () => {
+  if (notificationsCount === 0) return;
+  State.update({ open: true });
+}
+
 if (context.loading || !accountId) return <></>;
 
 const filterUsersRaw = Social.get(
@@ -32,7 +37,7 @@ if (filterUsers === null) {
 
 const filterUsers = filterUsersRaw ? JSON.parse(filterUsersRaw) : [];
 const notificationFeedSrc =
-  "${REPL_ACCOUNT}/widget/NearOrg.Notifications.Notifications";
+  "${REPL_ACCOUNT}/widget/NearOrg.Notifications.NotificationsList";
 const lastBlockHeight = Storage.get("lastBlockHeight", notificationFeedSrc);
 let notifications =
   Social.index("notify", accountId, {
@@ -78,8 +83,9 @@ const PreviewWrapper = styled.div`
   top: 70px;
   right: 68%;
   width: 460px;
+  max-height: 80vh;
   visibility: hidden;
-  overflow: hidden;
+  overflow: hidden auto;
   transition: visibility 300ms ease;
   transform-origin: right top;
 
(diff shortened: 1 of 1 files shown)


src/NearOrg/Notifications/NotificationButton.jsx
```diff
@@ -32,7 +37,7 @@ if (filterUsers === null) {
 
 const filterUsers = filterUsersRaw ? JSON.parse(filterUsersRaw) : [];
 const notificationFeedSrc =
-  "${REPL_ACCOUNT}/widget/NearOrg.Notifications.Notifications";
+  "${REPL_ACCOUNT}/widget/NearOrg.Notifications.NotificationsList";
```
shelegdmitriy, line 40: This is actually why notification counter got broken. In this [file](https://github.com/near/near-discovery-components/blob/develop/src/NearOrg/Notifications/NotificationsList.jsx#L21) we are making `Storage.set("lastBlockHeight", item.blockHeight)`. We've been trying to get a `lastBlockHeight` (line 41) from the given source component but that variable was in another component.

src/NearOrg/Notifications/NotificationButton.jsx
```diff
@@ -78,8 +83,9 @@ const PreviewWrapper = styled.div`
   top: 70px;
   right: 68%;
   width: 460px;
+  max-height: 80vh;
   visibility: hidden;
-  overflow: hidden;
+  overflow: hidden auto;
```
shelegdmitriy, line 88: This lines allows to scroll down the preview dropdown

src/NearOrg/Notifications/NotificationButton.jsx
```diff
@@ -15,6 +15,11 @@ State.init({
   open: false,
 });
 
+const handleDropdownOpen = () => {
+  if (notificationsCount === 0) return;
```
shelegdmitriy, line 19: As far as notification counter is working fine maybe it's good to hide preview dropdown is there are no new notifications. If you don't like this approach I'll remove it

cc @marcinbodnar @charleslavon

charleslavon: That's a good question @shelegdmitriy. @ravigautham should we change the UX such that if you have no unread notifications when you click the alert icon we show an empty box or some other affordance?

@shelegdmitriy I don't think we'll want to take no action upon clicking, because that might lead one to think that the UI is broken. Let's remove this for now and handle this potential UX improvement separately.

shelegdmitriy: Ok, I'll remove this. Let me explain more how I understand this approach (and how it's implemented here).

When user has some notifications - everything works as before. But when there are no any we are not showing dropdown menu but you can still click on bell button and see all the notifications on it's own page (like in our old implementation).

This is just my proposal. Totally agree that it could be improved in future and it's out of scope of this task.

charleslavon: Ah, I understand it much better now. Thanks @shelegdmitriy. That sounds like a worthwhile improvement to include on another PR.

charleslavon (approved):

PR #367, opened 2023-09-28
Prepare iOS Home Page prompt

shelegdmitriy:
Closes [#604](https://github.com/near/near-discovery/issues/604)

### DIFF
--- a/src/NearOrg/Notifications/HomeScreenAlert.jsx
+++ b/src/NearOrg/Notifications/HomeScreenAlert.jsx
@@ -1,0 +1,78 @@
+let { open, onOpenChange } = props;
+
+const ContentWrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
+  gap: 16px;
+`;
+
+const ImageWrapper = styled.div`
+  margin-top: -24px;
+`;
+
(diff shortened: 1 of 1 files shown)


charleslavon:
Hey @shelegdmitriy how do you suggest we test this?  I've tried with bos-loader, but nothing is rendered and no errors are thrown at `https://beta.near.org/test/widget/NearOrg.Notifications.HomeScreenAlert`. 

Also, this is the new modal and we'll need the logic which triggers this modal before showing the Enable Notifications modal. Does that already exists somewhere?

shelegdmitriy:
> Also, this is the new modal and we'll need the logic which triggers this modal before showing the Enable Notifications modal. Does that already exists somewhere?

Yes, you're right. We have to trigger this modal from `near-discovery` repository as we do for `NotificationAlert` component. So far there is no such logic. But as far as this [PR](https://github.com/near/near-discovery/pull/615) is merged I can use `isIOS` variable to create such functionality for iOS devices and there we can test it.

shelegdmitriy:
Actually we can make a pause with merging it and let me create such logic on `near-discovery` side and then make a loom video with an example. @charleslavon WDYT?

charleslavon (approved):

Issue #325, opened 2023-09-13
[bug] "All" and "Following" Toggles not Working on Activity Feed


ravigautham:
**Describe the bug**
Toggling between the "All" and "Following" filters on the activity feed is not rendering posts correctly. When I toggle to "Following" I still see content from all users and sometimes vice versa.

**To Reproduce**
1. Go to the activity feed
2. Click on "Following"
3. View Posts are incorrectly rendered

**Expected behavior**
- If a user has toggled "All" they should see posts from all users on near.org
- If a user has toggled "Following" they should see posts from only the users they follow

**Screenshots**
<img width="1512" alt="Screenshot 2023-09-13 at 2 47 07 PM" src="https://github.com/near/near-discovery-components/assets/129213007/fe1ec4a2-4766-423b-8e99-bd2658a81469">
<img width="1512" alt="Screenshot 2023-09-13 at 2 47 12 PM" src="https://github.com/near/near-discovery-components/assets/129213007/43d8792a-ce63-4a27-b9b6-4e8595074936">

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Arc Browser
 - Version: Version 1.7.0 (41350)

**Smartphone (please complete the following information):**

**Additional context**
Multiple reports of this bug from other users.

charleslavon:
@gabehamilton you might have some helpful insight on this issue if it's queryApi related

PR #323, opened 2023-09-13
Add notification preview

shelegdmitriy:
Closes: https://pagodaplatform.atlassian.net/browse/DEC-1431

Example of usage:

```jsx
<Widget
  src="discom.testnet/widget/NearOrg.Notifications.NotificationButton"
  props={{
    previewItems: <NotificationsList />,
    showMoreButton: <ShowMore />,
    notificationFeedSrc: components.nearOrg.notifications.page;
  }}
/>
```

PR #369, opened 2023-09-28
Fix `all` and `following` toggles on ActivityPage

shelegdmitriy:
Closes [#323](https://github.com/near/near-discovery-components/issues/325)

### DIFF
--- a/src/Posts.jsx
+++ b/src/Posts.jsx
@@ -1,6 +1,6 @@
 const GRAPHQL_ENDPOINT =
   props.GRAPHQL_ENDPOINT || "https://near-queryapi.api.pagoda.co";
-const LIMIT = 10;
+const LIMIT = 25;
 const accountsFollowing = props.accountsFollowing;
 const moderatorAccount = props?.moderatorAccount || "bosmod.near";
 const sort = Storage.get(`queryapi:feed-sort`) ?? "timedesc";
--- a/src/Posts.jsx
+++ b/src/Posts.jsx
@@ -1,25 +1,20 @@
 const GRAPHQL_ENDPOINT =
   props.GRAPHQL_ENDPOINT || "https://near-queryapi.api.pagoda.co";
-const LIMIT = 25;
-let accountsFollowing = props.accountsFollowing;
+const LIMIT = 10;
+const accountsFollowing = props.accountsFollowing;
 const moderatorAccount = props?.moderatorAccount || "bosmod.near";
 const sort = Storage.get(`queryapi:feed-sort`) ?? "timedesc";
+const initialSelectedTab = Storage.privateGet("selectedTab") ?? "all";
 
 State.init({
-  selectedTab: Storage.privateGet("selectedTab") || "all",
+  selectedTab: initialSelectedTab,
   posts: [],
   postsCountLeft: 0,
   initLoadPosts: false,
   initLoadPostsAll: false,
-  sort
+  sort,
+  accountsFollowing,
 });
-
-const previousSelectedTab = Storage.privateGet("selectedTab");
-if (previousSelectedTab && previousSelectedTab !== state.selectedTab) {
-  State.update({
-    selectedTab: previousSelectedTab,
-  });
-}
 
 const optionsMap = {
   timedesc: "Most Recent",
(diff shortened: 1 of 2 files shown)


src/Posts.jsx
```diff
@@ -1,54 +1,49 @@
 const GRAPHQL_ENDPOINT =
   props.GRAPHQL_ENDPOINT || "https://near-queryapi.api.pagoda.co";
-const LIMIT = 25;
-let accountsFollowing = props.accountsFollowing;
+const LIMIT = 10;
```
shelegdmitriy, line 3: Let's decrease `LIMIT` from 25 to 10? I don't see any reason why do we need to load 25 posts in fact that on 1 page could possibly fit something about 10 or even 8 posts.

src/Posts.jsx
```diff
@@ -210,16 +207,12 @@ query GetFollowingPosts($offset: Int, $limit: Int) {
 
 const loadMorePosts = () => {
   const queryName =
-    state.selectedTab == "following" && accountsFollowing
-      ? "GetFollowingPosts"
-      : "GetPostsQuery";
-  const type =
-    state.selectedTab == "following" && accountsFollowing ? "following" : "all";
+    state.selectedTab == "following" ? "GetFollowingPosts" : "GetPostsQuery";
```
shelegdmitriy, line 210: I don't really like that user without followed accounts will see the same posts from **"all"** tab but viewing **"following"** tab. It's confusing. This string enables filtering according to selected tab

src/Posts.jsx
```diff
@@ -210,16 +207,12 @@ query GetFollowingPosts($offset: Int, $limit: Int) {
...
+  const type = state.selectedTab;
 
   if (
     state.selectedTab == "following" &&
-    accountsSelected &&
-    accountsSelected.length == 0
```
shelegdmitriy, line 222: This two properties wasn't even declared. Seems like refactoring error or so. I guess it means `accountsFollowing` here. At least everything works fine

src/Posts.jsx
```diff
@@ -241,7 +234,7 @@ const loadMorePosts = () => {
           let filteredPosts = newPosts.filter((i) => !shouldFilter(i));
           filteredPosts = filteredPosts.map((post) => {
             const prevComments = post.comments;
-            const filteredComments = post.comments.filter(
+            const filteredComments = prevComments.filter(
```
shelegdmitriy, line 237: `prevComments` from 236 line was declared but not used so nothing changed here.

src/Posts/Feed.jsx
```diff
@@ -34,6 +34,14 @@ const renderItem = (item) => {
   );
 };
 
+if (posts.length === 0) {
```
shelegdmitriy, line 37: As from now we are more specific with filtering posts I decided to add this placeholder for users who doesn't have any posts from follow accounts

calebjacob (approved):

src/Posts.jsx
charleslavon: @shelegdmitriy how does this change the UX?

shelegdmitriy: Take a look at this [comment](https://github.com/near/near-discovery-components/pull/369#discussion_r1340735781) (the last one). If user doesn't have any followed account we'll show a message **"There is no any posts yet."**

charleslavon: @TiffanyGYJ  We could use this as a an opportunity to point the user to where they can find people/communitites to follow.    Could you suggest some text and a link we can add here?   Something like,  "[Connect](https://near.org/people) and find people to follow to customize your activity feed."

src/Posts.jsx
charleslavon: @shelegdmitriy please move this UX change and the "There are not any posts yet" change into separate PRs so we can get feedback from other people on those changes without holding up the fix in this PR.

shelegdmitriy: Sure I can revert `LIMIT = 25` but If I'll remove "There are not any posts yet" placeholder then **"Following"** tab will be empty. If I'll revert filtering logic (I mean we'll show posts from "all" tab on "following") then this PR doesn't make any sence. How about compromise? Let's add a line under "There are not any posts yet" with some text which will point to https://near.org/near/widget/PeoplePage ?

cc @charleslavon

shelegdmitriy: Like you proposed here https://github.com/near/near-discovery-components/pull/369#discussion_r1342728857

charleslavon: Good call @shelegdmitriy. We can iterate from there with feedback from @TiffanyGYJ

charleslavon: Instead of "There are not any posts yet" please use "Build your feed by finding [people to follow](https://near.org/near/widget/PeoplePage)"

calebjacob (approved):

PR #380, opened 2023-10-04
Fix new notification counter

shelegdmitriy:
Potentially closes [#655](https://github.com/near/near-discovery/issues/655)


shelegdmitriy:
I mention that before this PR (and on prod) the notifications counter shows the right number of new notifications and disappears in second or less. I checked that on near.social and NotificationButton works fine and displays the right number of new notifications. So I decided to refactor the whole component using the same component from near.social. Here is how this button looks like: [near.social/NotificationButton](https://near.social/mob.near/widget/WidgetSource?src=mob.near/widget/NotificationButton)

calebjacob (approved):

PR #381, opened 2023-10-04
Fix new notification counter & cleanup

shelegdmitriy:
Closes [#655](https://github.com/near/near-discovery/issues/655)


src/NearOrg/Notifications/NotificationButton.jsx
```diff
@@ -133,37 +133,40 @@ const Notification = ({ count, disabled }) => {
...
-              getNotificationLocalStorage,
-              handleOnCancelBanner,
-              accountId,
-              handleTurnOn,
-              showLimit: 5,
-              showInBox: true,
-            }}
-          />
-          <SeeAll>
+        {state.open && (
```
shelegdmitriy, line 136: Actually this line is a solution for counter. All other changes just clean up

shelegdmitriy: Let me explain why:

There is a `NotificationsList` component inside `Notifications`. There (in `NotificationsList`) we are calling `Storage.set("lastBlockHeight", item.blockHeight)` ([code](https://github.com/near/near-discovery-components/blob/develop/src/NearOrg/Notifications/NotificationsList.jsx#L21))

```
if (i === 0) {
  Storage.set("lastBlockHeight", item.blockHeight);
}
```

and `Storage.set("lastBlockHeight", item.blockHeight)` is out starting point to query for new notifications.

Let's get back to `NotificationButton` component. Without `state.open` check we are sending queries to check new notifications twice:

- first query return the right number of it (that's why there was a quick flash with notification number in first second of load);
- second query returns an empty array because first query already did that and `"lastBlockHeight"` is not the latest no more.

I hope this helps 😀

calebjacob: Thanks for the explanation! 👍

calebjacob (approved):

Issue #360, opened 2023-09-26
Implement new near.org home page design


calebjacob:
https://www.figma.com/file/50VvVImqAoHAhJctYdPBzv/%F0%9F%9B%A0%EF%B8%8F-Near.org---v2.2023?type=design&node-id=562-68289&mode=design&t=Y9SvaMa6RNWRpmVP-0

After discussing with @charleslavon, we agreed to hard code the latest events and news links from the NF WordPress website. We can pull these in dynamically as an enhancement later by setting up a Next JS `/api` route on our gateway that crawls the WordPress website.

PR #385, opened 2023-10-06
New Home Page

calebjacob:
Closes: https://github.com/near/near-discovery-components/issues/360

We can safely merge and publish this component without actually affecting `near.org`. The gateway code still uses our hard coded Next JS pages from when we thought we had an SEO issue. Once all the new page BOS components are finished, I'll submit a gateway PR to render the new BOS components.

One change I made in this PR is removing all the CSS variables from `DIG.Theme` to avoid confusion and duplication. Our discovery gateway has the same values copied into `theme.css`, which already provides these variables to the entire Next JS app and BOS components on our gateway. This `theme.css` file is necessary at the Next JS level so our primary navigation components (which are native React/Next components) have access to our theme colors. We can keep thinking through the best way to handle theme colors going forward, but at least for now, this change will help reduce confusion. Let me know if you have any concerns!


src/NearOrg/LatestNews.jsx
```diff
@@ -0,0 +1,25 @@
+/*
+  The following data fetching was pulled from this source code:
```
charleslavon, line 2: 🔥

src/NearOrg/TotalStats.jsx
```diff
@@ -0,0 +1,24 @@
+const stats = {
+  totalApps: 0,
+  totalComponents: 0,
+  totalDevelopers: 0,
+};
+const componentsData = Social.keys("*/widget/*", "final", {
```
charleslavon, line 6: nice approach to making this dynamic

charleslavon:
Do we have a reasonable option to improve the centering of these numbers over their labels?
Screen Shot 2023-10-06 at 4 17 27 PM

charleslavon:
awesome work @calebjacob this looks super good

charleslavon (approved):

calebjacob:
> Do we have a reasonable option to improve the centering of these numbers over their labels? Screen Shot 2023-10-06 at 4 17 27 PM

Good point! I pushed a commit to center those labels on the numbers. Thanks for the feedback! :)

PR #396, opened 2023-10-10
Reference copy of the social_feed indexer

gabehamilton:
This is a reference copy of the moderation indexer. The running code at https://near.org/dataplatform.near/widget/QueryApi.App?selectedIndexerPath=dataplatform.near/social_feed&view=editor-window is the source of truth.


roshaans (approved):
LGTM

roshaans:
I think I would call the files `index.js` and `schema.sql` but this is just my personal preference.

PR #397, opened 2023-10-10
Reference copy of the verifications indexer

gabehamilton:
This is a reference copy of the moderation indexer. The running code at https://near.org/dataplatform.near/widget/QueryApi.App?selectedIndexerPath=dataplatform.near/verifications&view=editor-window is the source of truth.


replacements.testnet.json
```diff
@@ -1,5 +1,4 @@
 {
-  "REPL_ACCOUNT": "discom.testnet",
```
roshaans, line 2: uninteded change

Issue #402, opened 2023-10-12
Extra whitespace after mentions


gabehamilton:
**Describe the bug**
When an account is mentioned in a post, extra whitespace is appearing after it.

![Screenshot 2023-10-12 at 11 15 54 AM](https://github.com/near/near-discovery-components/assets/671506/daa9463a-b4cf-4c78-8fb7-17f3e25fc2e1)

PR #403, opened 2023-10-12
Render mentions inline

gabehamilton:
See [bug](https://github.com/near/near-discovery-components/issues/402) for screenshot of problem. 

Add span from https://near.social/mob.near/widget/WidgetSource?src=mob.near/widget/SocialMarkdown so mentions are inline. 

Should this instead be applied in the AccountProfileInline component?
Reviewers, feel free to commit a different solution to this branch if this span is not the best solution.

### DIFF
--- a/src/SocialMarkdown.jsx
+++ b/src/SocialMarkdown.jsx
@@ -54,6 +54,7 @@ const Wrapper = styled.div`
 const renderMention =
   props.renderMention ??
   ((accountId) => (
+  <span className="d-inline-flex">
     <Widget
       key={accountId}
       src="${REPL_ACCOUNT}/widget/AccountProfileInline"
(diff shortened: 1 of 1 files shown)


calebjacob (changes requested):
This is an interesting bug. I'm afraid your change won't really address the issue @gabehamilton - and it will most likely break our text overflow handling for super long usernames.

The real culprit is that we're using `white-space: pre-line` on `<p>` tags inside the markdown section. Check out all the line breaks inside a `<p>` tag that uses a mention:

Screen Shot 2023-10-12 at 11 47 08 AM

We need to somehow remove the line breaks. I also noticed there's a new element that renders after every mention now:

```
<div role="region" aria-label="Notifications (F8)" tabindex="-1" style="pointer-events: none;">...</div>
```

This needs to use `display: none;` or `position: absolute;` - it's currently also adding whitespace.

calebjacob:
Hey @gabehamilton, check out this PR: https://github.com/near/near-discovery-components/pull/404

PR #418, opened 2023-10-16
chore: fix alignment issue

charleslavon:
@shelegdmitriy or @calebjacob my fix is probably the sloppiest approach. Would you suggest an alternative?

before: 
Screen Shot 2023-10-16 at 3 54 16 PM

with this PR (ignore the super wide display):
Screen Shot 2023-10-16 at 3 54 03 PM


shelegdmitriy:
@charleslavon, sorry I'm out of context, what are you trying to fix?

shelegdmitriy (approved):

PR #423, opened 2023-10-17
Non-regular promotion of develop to main

shelegdmitriy:
This update is initiated in favor to promote  #422 into production but it seems we can make a prod release.

cc @charleslavon @calebjacob any objections?


charleslavon:
@gabehamilton does this PR included the activityfeed change that introduced the error which we ended up reverting?

gabehamilton:
@charleslavon it has the fixed version of that. Please check the feed after deployment, though :-)

PR #424, opened 2023-10-17
375 filter activityfeed based on new moderation format

gabehamilton:
Fixed version of Filter ActivityFeed based on new moderation format.

This is the same as what was merged yesterday with the addition of null checks on the Social.getr method when fetching selfModeration data and the same for selfFlagged data. 

Also switched the fallback/lag check in ActivityPage to use the dataplatform_near_moderated_posts table for consistency.

### DIFF
--- a/src/Posts.jsx
+++ b/src/Posts.jsx
@@ -26,7 +26,7 @@ const queryFollowedAccountsList = () => {
   const graph = Social.keys(`${context.accountId}/graph/follow/*`, "final");
   if (graph !== null) {
     const followedAccounts = Object.keys(
-      graph[context.accountId].graph.follow || {}
+      graph[context.accountId].graph.follow || {},
     );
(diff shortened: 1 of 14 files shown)


src/Posts.jsx
```diff
@@ -45,68 +45,50 @@ const selectTab = (selectedTab) => {
...
+//         ".post.main": { // slashes are not allowed in keys
+//           "100000123": "spam", // post ids are account/blockHeight
+//         }
 //     },
 //   }
 // }
-function matchesModeration(moderated, item) {
-  let accountFound = moderated[accountId];
+const selfModeration = context.accountId
+  ? Social.getr(`${context.accountId}/moderate`, "optimistic") ?? []
```
gabehamilton, line 67: The `?? []` is the change that fixes the feed not populating. Applied the same to flaggedPosts just to be sure.

PR #428, opened 2023-10-18
feat: Create a shorthand command to run bos-loader

shelegdmitriy:
As far as I need to start my local env quite ofter I decided to create a shorthand for starting this project with bos-loader. This code is not really my, I created it using Chart-GPT and it seems to work very well 🙂

To start using it you need to do this once
```
chmod +x start.sh
```

And then from home directory run this command:

```
./start.sh testnet discom.testnet
```

At the end you're still need to revert `replacements.*.json` changes.

Feel free to leave your comments here

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+
+# Store the current working directory
+CURRENT_DIR=$(pwd)
+
+# Check if environment and account were provided as arguments
+if [ "$#" -ne 2 ]; then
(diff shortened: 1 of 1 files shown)


mpeterdev (commented):
nice, I love helper scripts like these.

it already generates a temp replacements file, you could have it just use that in the bos-loader command instead of overwriting the committed replacements file and would then not need to revert changes to it before committing

happy to approve as is or wait on that as an improvement

Issue #425, opened 2023-10-17
Reduce number of HTTP calls on ActivityPage


calebjacob:
The `ActivityPage` currently executes around 500 HTTP requests before it finishes loading. This is a ridiculous amount of requests and is a massive bottleneck for even modern browsers and high speed internet. On lower end devices or slower internet connections, this makes it almost unusable.

A lot of these requests are coming from nested components using things `Social.get()` and the `TimeAgo` component. We need to batch these requests or rely on Query API to provide this data as much as possible.

charleslavon:
@calebjacob is this 500 request evaluation based on [the QueryAPI feed](https://github.com/near/near-discovery-components/blob/b64b389ea05fbbfc7c65c42026b62d58eabeceb3/src/Posts/Feed.jsx#L4). This feed has a trigger which defaults to the [previous Social.get() feed](https://github.com/near/near-discovery-components/blob/b64b389ea05fbbfc7c65c42026b62d58eabeceb3/src/v1/Posts.jsx#L1) in the event of problems with QueryAPI. 

@gabehamilton have I described that correctly?

gabehamilton:
Correct if the blockheight difference between the QueryApi table the Social.index posts is greater than 2 it falls back. I think I've been seeing about 150 requests in the QueryApi version, plus plenty more with every scroll. Still definitely a problem, issue description is spot on in general.

Issue #427, opened 2023-10-18
Feeds should pull in profile fields via graphql relationships


gabehamilton:
To improve the performance of the near.org feeds and reduce the number of data requests, the feed queries can fetch profile fields through new relationships to the accounts indexer.

Note, this Issue depends on #612 where the moderated_posts View moves into the `feed` indexer.

1. Create a hasura relationship between social_feed.posts and accounts.accounts tables. Create one between feed.moderated_posts and accounts.accounts. Determine whether to use foreign keys to create the relationships in the `feed` indexer's sql or to invoke hasura stored procedures.
2. Update the feed queries in PostsFeedControls.jsx to fetch profile fields: profile_name, profile_image, profile_tags using the account relationship.
3. Pass the profile fields through to the AccountProfile component (profile prop).

Note: Title and description updated Jan 16 2024

charleslavon:
@shelegdmitriy is related to [885](https://github.com/near/near-discovery/issues/855)?

shelegdmitriy:
> @shelegdmitriy is related to [885](https://github.com/near/near-discovery/issues/855)?

Not sure but could be... @gabehamilton could you please clarify this a bit?

gabehamilton:
Yes, it's related to #885 and I'll rewrite the title and description of this ticket to match what we should do. This Issue got the notes about what fields we needed but we soon realized we needed the full account indexer. Now that we have that we can add relationships between social_feed.posts & accounts.accounts and feed.moderated_posts and accounts.accounts. That will allow the feed queries in PostsFeedControls to pull in the profile fields through the relationship.

PR #429, opened 2023-10-19
Improved ActivityPage Performance

calebjacob:
Closes: https://github.com/near/near-discovery-components/issues/425

This is a first rough pass at fixing some performance issues. I was mainly focused on quick wins that wouldn't require Query API changes. We can work with @gabehamilton in future tickets to enhance our Query API data to continue reducing things like cascading `Social.get()` calls. I was able to eliminate about 130 HTTP requests in this PR, but we still have a massive amount of room for improvement.

I also fixed the flickering that was occurring when hovering over a user's profile name/avatar (`AccountProfileOverlay`). I refactored this to use Radix's `Popover` component - which doesn't require changing state and causing further re-renders.

The main activity feed was fetching data multiple times and causing expensive re-renders. I refactored the logic that loads posts to avoid all of these duplicate requests on page load.

I noticed that Eugene's `TimeAgo` component always makes an HTTP call to calculate a timestamp on the backend. I created a new `TimeAgo` component that can utilize `block_timestamp` passed down by Query API to parse the timestamp entirely on the frontend and avoid the HTTP call. It will fall back to Eugene's `TimeAgo` component if the timestamp isn't passed through.

These external components rendered in the sidebar could really use a refactor as well to reduce HTTP calls:

- nearweekapp.near/widget/NEARWEEKNews
- nearweekapp.near/widget/nearweek-newsletter

Please test these changes locally if you can @charleslavon @gabehamilton @shelegdmitriy. I've done my best to ensure I didn't break anything, but always could use more eyes. Would be great to get this merged ASAP to avoid really ugly merge conflicts. :)

Also, running the activity page locally with `pnpm run dev` is still pretty laggy, but it's definitely improved. The UI is usable if you wait a few seconds - which is some improvement.


src/Posts/Feed.jsx
```diff
@@ -27,26 +51,25 @@ const renderItem = (item) => {
...
           accountId: item.account_id,
           blockHeight: item.block_height,
+          blockTimestamp: item.block_timestamp,
           content: item.content,
           comments: item.comments,
           likes: item.accounts_liked,
           GRAPHQL_ENDPOINT,
           verifications: item.verifications,
           showFlagAccountFeature: props.showFlagAccountFeature ?? false,
+          profile: returnProfileForUser(item),
```
charleslavon, line 61: I assume passing this saves some queries within Posts.Post?

src/Recommender/Engagement/ImageTracked.jsx
```diff
@@ -18,14 +25,14 @@ return (
...
-          image: props.profileImage,
-          alt: props.profileImageAlt,
-          fallbackUrl:
-            "https://ipfs.near.social/ipfs/bafkreibiyqabm3kl24gcb2oegb7pmwdi6wwrpui62iwb44l7uomnn3lhbi",
-        }}
-      />
+    <Widget
+      src="${REPL_MOB}/widget/Image"
+      props={{
+        image,
```
charleslavon, line 31: is the intent here to create this object only once, instead of on each render?

src/LatestFollowActivity.jsx
```diff
@@ -56,28 +71,37 @@ return (
...
-          />
-        </Item>
-      ))}
+            <Text small bold>
+              <i className="bi bi-arrow-right"></i>
+              {/* <br />
+              <Widget
+                src="${REPL_MOB_2}/widget/TimeAgo${REPL_TIME_AGO_VERSION}"
+                props={{ blockHeight: item.blockHeight }}
+              /> */}
```
shelegdmitriy, line 92: Why are this lines commented?

charleslavon (approved):
looks awesome. let's test on staging

src/Posts/Feed.jsx
calebjacob: Correct. I chatted with @gabehamilton, and his plan is to pass the profile info of each author with Query API. This will prevent us having to call `Social.get()` inside `AccountProfile` to fetch the user's display name and avatar.

calebjacob: https://github.com/near/near-discovery-components/issues/427

src/Recommender/Engagement/ImageTracked.jsx
calebjacob: This logic was needed to create the correct `image` object for the `Image` component based on the value of `props.profileImage`. Sometimes it's a full URL and other times it's an IPFS CID.

shelegdmitriy (commented):
Looks great actually! Need to check it locally but @calebjacob you did a great job 👏

src/LatestFollowActivity.jsx
calebjacob: Good catch - I forgot to add this back. We don't currently have any way to avoid the HTTP requests here for rendering the follow timestamp, but it should only create 5 extra requests. This would be an example where Query API can eventually help out.

PR #439, opened 2023-10-24
Needs Moderation interface shows reported accounts posts comments

gabehamilton:
Note, if you test this locally, it only works against `mainnet` due to the dependency on QueryApi. To fully test end to end with a user other than `bosmod.near` reach out to Gabe.

## New Features
Moderator
 * Moderator interface has sidebar for Overview, Needs Moderation, Previously Moderated.
 * Previously Moderated interface allows unmoderation of content and manual moderation.
 * Needs Moderation interface allows moderators to work with feed of reported, not yet moderated content.

Needs Moderation
 * Lists Reported Accounts, Posts, Comments in one feed, grouped by each reported item.
 * Content is hidden until expanded. Meta info in header and header tooltips.
 * Posts and Comments can be individually blocked or whole account can be moderated.
Screenshot 2023-10-24 at 12 52 33 PM
 * Expanded content shows account, post, or comment
Screenshot 2023-10-24 at 12 55 32 PM
 * Moderated content gets local 'Moderated' message until list is reloaded.
Screenshot 2023-10-24 at 12 58 24 PM

### DIFF
--- a/src/Moderation/Sidebar.jsx
+++ b/src/Moderation/Sidebar.jsx
@@ -1,0 +1,68 @@
+const Wrapper = styled.div`
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  border-right: 1px solid #eceef0;
+  gap: 24px;
+  width: 240px;
+  height: 100%;
+`;
+
+const Title = styled.h2`
+  font: var(text-l);
+  font-weight: 600;
+`;
+
+const MenuItem = styled.div`
+  display: flex;
+  align-items: center;
+  gap: 10px;
+  cursor: pointer;
+  transition: 0.2s all;
+  font-weight: 400;
+  ${(p) =>
+    p.active &&
+    `
+    font-weight: 700;
+  `}
+`;
+
+const Text = styled.span`
+  font: var(--text-s);
+  font-weight: inherit;
+  line-height: 16px;
+`;
+
+const Icon = styled.i`
+  font-size: 24px;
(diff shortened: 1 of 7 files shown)


src/Moderation/Sidebar.jsx
```diff
@@ -0,0 +1,74 @@
...
+  width: 240px;
+  height: 100%;
+`;
+
+const Title = styled.h2`
+  font: var(text-l);
+  font-weight: 600;
+`;
+
+const MenuItem = styled.div`
```
calebjacob, line 16: For the short term, let's at least make `MenuItem` a `<button>` element. That way it will be more accessible. You might need to review the styles after switching from `div` to `button`. Often times, adding `all: unset;` to your button styles can be helpful to reset any inherited styles.

calebjacob (changes requested):
Great work using the `DIG` components! See my one comment about `MenuItem` in `Sidebar`.

In the future, it might be worth refactoring the `Sidebar` to rely on `a` tags to set the current moderation page via URL. See `ActivityPage` as an example. The benefit here is that you could link directly to any of the moderation pages in the `Sidebar`. You don't have to do that in this PR, but maybe in a separate ticket.

src/Moderation/Sidebar.jsx
gabehamilton: Nice! Done.

calebjacob (approved):

PR #441, opened 2023-10-24
Fix/recommenders handle 404 with message

scopalaffairs:
This PR introduces 

- fixed behavior for a UI message on errors, e.g. 404.
- streamlined function `getRecommendedUsers` with corrected state handling to use asyncFetch only.
- fixed behavior for low metric user data, e.g. likes, followers, following when either of them should be null.
- removed/fixed EngagementTracker response handling, only logging errors, since OK is ok.
- additional functionality to get next users onFollow click in ActivityPage > LatestPeople > Recommended: followed user is removed, subsequent user is added.
- general minor cleanups


calebjacob:
Hey @scopalaffairs, thanks for the PR! I recently refactored some of the same components to improve performance and fix a couple bugs along the way. Unfortunately that means there's a big merge conflict now. Can you pull from `develop` and review the updated code? Then you can re-apply any of your changes that you still feel are needed. Thanks!

scopalaffairs:
Hey @calebjacob, thanks for the heads up and the good work there.
Main thing I had to resolve is how the `returnProfileForUser` function you introduced lately is called.
It looks good on my side, PeoplePage and especially the ActivityPage > LatestPeople > hover on profiles, that's very much improved and smooth now! Great!

scopalaffairs:
@calebjacob, `returnProfileForUser` would also make sense to have on the other components that do the same.
Do you think, it's a good idea to add that in this PR?

src/Recommender/Views/TrendingUsersView.jsx
```diff
@@ -135,9 +141,10 @@ function returnProfileForUser(user) {
   }
 
   return null;
-}
+};
 
-if (state.isLoading && !state.hasFetched) {
+if (!state.isLoading && !state.hasLoaded) {
+  returnProfileForUser();
```
calebjacob, line 147: @scopalaffairs Just curious why we need to call `returnProfileForUser()` here? It doesn't seem to have any affect.

scopalaffairs: You're right, that mistake might have been related to the merge conflict resolving. 
Thanks for catching that!

src/Recommender/Service/RecommendedUsers.jsx
```diff
@@ -59,82 +59,101 @@ const RecommendedUsers = styled.div`
   padding-bottom: ${props.sidebar ? "12px" : "50px"};
 `;
 
+const NotEnoughData = styled.div`
+  width: 100vw;
```
calebjacob, line 63: Sorry I didn't catch this earlier in my review. We should probably be using `width: 100%` to fill the container instead of `100vw`. This component is used in the `ActivityPage` sidebar and using `100vw` would break our layout (it would make the sidebar as wide as the screen).

calebjacob (changes requested):

src/Recommender/Service/RecommendedUsers.jsx
scopalaffairs: 👍

scopalaffairs:
hey @calebjacob, sorry for making any post-approval changes.
I had the branch tested with @TiffanyGYJ  and we adjusted the TrendingUsersView error message to make more sense.
Also, I reverted the width for the components in PeoplePage to be 100vw again, since 100% does not use the full space and looks weird.

calebjacob (approved):
No worries. Looks good.

calebjacob (approved):

PR #442, opened 2023-10-24
Fix Activity Page Flashing (Part 1 of 2)

calebjacob:
Related to: https://github.com/near/near-discovery/issues/673

Once this PR is merged, I'll open another PR on the gateway repo for the 2nd half of the fix.

- Break out all activity feed fetch logic into its own component to avoid re-rendering ActivityPage
- Add `GatewayWrapper` to standardize usage of `TosCheck` and make it easier to add global functionality in the future (this will be used by the gateway in the additional PR). This also prevents a re-render of the entire page once the `tosData` loads.


charleslavon (approved):
LGTM

PR #449, opened 2023-10-25
Optimistically hide post body when hiding or reporting.

gabehamilton:


gabehamilton:
The comment optimistic hiding is not working, however I think it's good to keep those changes in so comments stay in sync code wise with posts for the rest of the feature.

PR #452, opened 2023-10-26
Link Refactor

calebjacob:
Closes: https://github.com/near/near-discovery/issues/739

- Replaces all internal `<a>` tags with the `<Link>` component
- Replace all `#/...` URLs with `/...`
- Updated `DIG.Button` to internally rely on `Link` when passed an `href`

This `Link` component implements `next/link` so that proper client side routing is executed when clicked. If you're curious, this `Link` component is made globally available thanks to the VM: https://github.com/NearSocial/VM/releases/tag/2.1.0

Going forward, please use the `Link` component when creating an internal link and don't include a hash (`#/...`) in any `href`. If you need to set up an external link that opens a new tab, you can continue to use a `<a>` tag with the `target="_blank"` attribute.

I apologize ahead of time for any merge conflicts this might create since it touches so much of our codebase. :)

CC: @shelegdmitriy @gabehamilton @charleslavon @roshaans @scopalaffairs


shelegdmitriy (approved):
Great job @calebjacob 💪

shelegdmitriy:
I think we need to add an information about usage of `<Link>` tag vs `<a>` to the contributors file. Also about links to other components which should start from `/` and not `#/`. I think it's worth to mention that this two rules are very recommended to use in our gateway

charleslavon (approved):
I agree with Dima that docs would be helpful, but let's get this merged to avoid conflicts.

PR #466, opened 2023-10-31
Use ActivityFeed on ProfilePage

gabehamilton:
Main features
* By using the main Activity feed, standard moderation and self-moderation is applied to the content.
* Add account filter to Activity feed 
* Use Activity Feed on Profile Page with account filter
* Add Mutual Activity feed for use on profile page instead of Following Feed

Small adjustments
* Only show feed selector if there is more than one.
* Show sort selector for all feeds
* property determines whether to show Compose Post area.

This is a screenshot of the Overview portion of a user's profile (when viewed by a different user). 
Screenshot 2023-10-31 at 11 19 27 AM

### DIFF
--- a/src/ActivityFeeds/PostsFeedControls.jsx
+++ b/src/ActivityFeeds/PostsFeedControls.jsx
@@ -10,12 +10,23 @@ State.init({
 const GRAPHQL_ENDPOINT =
   props.GRAPHQL_ENDPOINT || "https://near-queryapi.api.pagoda.co";
 const LIMIT = 10;
-const moderatorAccount = props?.moderatorAccount || "bosmod.near";
+const feeds = props.feeds ?? ["all", "following"];
+const feedLabels = {all: "All", following: "Following", mutual: "Mutual Activity"};
+const showCompose = props.showCompose ?? true;
+const filteredAccountIds = props.filteredAccountIds;
+
 const sortRaw = Storage.get("queryapi:feed-sort");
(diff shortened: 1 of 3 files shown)


src/ActivityFeeds/PostsFeedControls.jsx
```diff
@@ -10,12 +10,23 @@ State.init({
 const GRAPHQL_ENDPOINT =
   props.GRAPHQL_ENDPOINT || "https://near-queryapi.api.pagoda.co";
 const LIMIT = 10;
-const moderatorAccount = props?.moderatorAccount || "bosmod.near";
```
gabehamilton, line 13: Should have been removed in recent moderation PRs, not in use.

src/ActivityFeeds/PostsFeedControls.jsx
```diff
@@ -105,54 +116,50 @@ const createQuery = (type) => {
...
     default:
       querySortOption = "";
   }
 
   let queryFilter = "";
   switch (type) {
     case "following":
+      let filteredAccountsFollowing = accountsFollowing;
+      if(filteredAccountIds) {
+        const filteredAccountList = filteredAccountIds.split(",");
```
gabehamilton, line 128: Nothing is using the filtering of the following list yet. I put it in for consistency in case groups or other future features both filter and use this feed.

calebjacob (approved):
Looks good!

PR #488, opened 2023-11-04
Add mobile view to the Settings page

shelegdmitriy:
This PR introduces a mobile view for Settings page. This is how it looks like:

Знімок екрана 2023-11-04 о 15 12 05
Знімок екрана 2023-11-04 о 15 12 16

I have hidden the tabs in the Sidebar because so far we only have 1 tab. If we add more tabs they will be shown automatically.

### DIFF
--- a/src/Settings/Sidebar.jsx
+++ b/src/Settings/Sidebar.jsx
@@ -10,12 +10,21 @@ const Wrapper = styled.div`
   gap: 24px;
   width: 264px;
   height: 100%;
+
+  @media (max-width: 1024px) {
+    border-right: none;
+    width: 100%;
+  }
 `;
 
 const Title = styled.h2`
-  font: var(text-l);
+  font: var(--text-l);
   font-weight: 600;
-  padding: 64px 24px 24px;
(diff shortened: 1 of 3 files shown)


src/Settings/Sidebar.jsx
```diff
@@ -80,7 +89,7 @@ return (
   <Wrapper>
     <Title>Settings</Title>
 
-    {filteredMenuItems.map((item) => (
+    {filteredMenuItems.length > 1 && filteredMenuItems.map((item) => (
```
charleslavon, line 92: @shelegdmitriy why only return a menuItem if there is > 1?

charleslavon: Ah, because it defaults to iDOS?

charleslavon (approved):

src/Settings/Sidebar.jsx
shelegdmitriy: Correct. At this moment we have only one tab so it doesn't make much sense to show it

PR #497, opened 2023-11-06
Latest social feed with additional try/catch in handleLike

gabehamilton:


indexers/social_feed/social_feed.js
```diff
@@ -166,70 +166,74 @@ async function getBlock(block: Block) {
...
+          // Comment
+          console.log(`Likes to comments are not supported yet. Skipping`);
+          break;
+        default:
+          // something else
+          console.log(`Got unsupported like type "${itemType}". Skipping...`);
+          break;
+      }
+    } catch (likeError) {
+      console.log("Error in handleLike", likeError);
```
darunrs, line 236: Can we instead say something like "Failed to process like acction content."? Since the rest of the code is in try catches, we know that this one should catch an error related to reading the json and preparing variables.

gabehamilton: I'll just close this in favor of your PR.

gabehamilton:
Content is covered in https://github.com/near/near-discovery-components/pull/496

PR #500, opened 2023-11-07
Removes the redirect to a custom ProfilePage component

charleslavon:
Support the latest ProfilePage features for all users on near.org by removing the redirect to a custom ProfilePage component


calebjacob:
I definitely agree with this change @charleslavon - this keeps tripping me up since I created a `ProfilePage` component way back for initial testing. There's no way to undo it, so my account has been permanently stuck on a stale profile page for awhile now.

gabehamilton (approved):

gabehamilton:
It is a cool feature. I think we just need to wrap it in tabs of some sort so that as the viewing user I can switch between the standard and custom profiles.

PR #502, opened 2023-11-07
Fix caching issue with component details stats

roshaans:
fix caching problem temporarily by redirecting redirecting user to new window which forces a re-render

### DIFF
--- a/src/ComponentProfile.jsx
+++ b/src/ComponentProfile.jsx
@@ -56,7 +56,7 @@ const Thumbnail = styled.div`
 `;
 
 return (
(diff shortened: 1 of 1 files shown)


calebjacob:
What caching issue is this trying to resolve? I'm afraid this change will lead to poor UX.

pkudinov:
If you open component details page and click through dependent components, the number of impressions doesn't change. If on any of those components you hard refresh the page, the correct number shows up.

This is pretty critical right now as in many cases we are showing incorrect data. We can work on the proper fix later

calebjacob:
Thanks for the background @pkudinov

calebjacob (approved):

PR #516, opened 2023-11-09
Check type of accounts_liked data before parsing.

gabehamilton:


calebjacob (approved):

src/Posts/Post.jsx
```diff
@@ -77,7 +77,9 @@ query IndexerQuery {
           const post = posts[0];
           let content = JSON.parse(post.content);
           if (post.accounts_liked.length !== 0) {
-            post.accounts_liked = JSON.parse(post.accounts_liked);
+            if(typeof post.accounts_liked === "string") {
```
morgsmccauley, line 80: Does this mean non-string values are ignored? The `accounts_liked` only seems to be JSON now, so this would ignore all new likes?

gabehamilton: In the recent data from the indexer the accounts_liked are coming through as a JSON array at this point thus don't need any treatment.

morgsmccauley: Right, I see, thanks for explaining :)

PR #517, opened 2023-11-09
Star Component MVP

calebjacob:
Closes: https://github.com/near/near-discovery/issues/692

After talking with @thisisjoshford, we decided to just implement an MVP version of the requirements in the ticket. We'll build out the new developer profile page at a later time, so we added a `Stars` tab to the current profile page.

I created a generic component for actions like favoriting or starring that we can use going forward: `SocialIndexActionButton`. I copied most of the logic from `mob.near`'s component and updated it a bit to make it more generic: https://near.org/near/widget/ComponentDetailsPage?src=mob.near/widget/N.StarButton

I also implemented a new notification type for the `star` action so you can be notified when someone stars your component (it won't notify you if you star your own component). While updating the notifications, I noticed there were some older notification components that we needed to flag as deprecated to avoid confusion.

I updated a few existing properties that were triggering errors in the console on the component details page too. EG: `style={{ 'margin-top': 10px }}` => `style={{ marginTop: 10px }}`. Also ended up applying some code formatting.


thisisjoshford:
sweeeet! tysm @calebjacob ! :)

src/SocialIndexActionButton.jsx
```diff
@@ -0,0 +1,89 @@
...
+  } else if (action.value.type === actionUndoName) {
+    delete actionsByUsers[action.accountId];
+  }
+});
+
+if (actionsByUsers[context.accountId] && isActive === null) {
+  setIsActive(true);
+}
+
+if (isActive !== null) {
```
gabehamilton, line 21: Could you explain this block? I don't understand its purpose since actionsByUsers was just built from the index.

gabehamilton (approved):

src/SocialIndexActionButton.jsx
calebjacob: Sure! The logic inside `if (isActive !== null) { ... }` updates `actionsByUsers` whenever the user clicks the button - which happens after the index data is initially fetched. So we need to update the stale data in order for the count to be updated instantly when evaluating `const count = Object.keys(actionsByUsers).length;`.

PR #520, opened 2023-11-13
Added block menu in notification list

erditkurteshiSQA:
Fixes [#776](https://github.com/near/near-discovery/issues/776).

Added block menu on notification list and also confirmation modal.

 Related with: https://github.com/near/near-discovery/pull/820.

block menu

image


src/NearOrg/Notifications/Notification.jsx
```diff
@@ -14,12 +14,15 @@ const Notification = styled.div`
   }
 `;
 
-const Content = styled.div`
+const Content = styled.a`
```
shelegdmitriy, line 17:
```suggestion
const Content = styled("Link")`
```

@erditkurteshiSQA check out this [discussion](https://github.com/near/near-discovery-components/discussions/498) for the reference

src/NearOrg/Notifications/Notification.jsx
```diff
@@ -322,5 +357,35 @@ return (
...
     </Content>
+    <Widget
+          src="${REPL_ACCOUNT}/widget/DIG.DropdownMenu"
+          props={{
+            trigger: <i className="ph-bold ph-dots-three" />,
+            items: buildMenu(),
+          }}
+        />
+        {isModalOpen && (
+        <div className="modal" style={{ display: 'block', backgroundColor: 'rgba(0, 0, 0, 0.6)' }}>
```
shelegdmitriy, line 368: For the modal we have a [DIG.Dialog](https://near.org/near/widget/ComponentDetailsPage?src=near/widget/DIG.Dialog&tab=about) component

shelegdmitriy (changes requested):
@erditkurteshiSQA sorry for reviewing the draft version of this PR but I left a couple of comments

src/NearOrg/Notifications/Notification.jsx
erditkurteshiSQA: @shelegdmitriy This has been addressed in the latest commits. Thank you for the review.

src/NearOrg/Notifications/Notification.jsx
erditkurteshiSQA: This has been also addressed in the latest commit. Thank you for the review.

src/NearOrg/Notifications/Notification.jsx
```diff
@@ -130,6 +133,7 @@ const path = item.path || "";
 
 // Build notification
 let { blockHeight, accountId } = props;
+let { manageNotification } = props;
```
shelegdmitriy, line 136:
```suggestion
let { blockHeight, accountId, manageNotification } = props;
```

src/NearOrg/Notifications/NotificationsList.jsx
```diff
@@ -16,6 +16,8 @@ const index = {
   },
 };
 
+let { manageNotification } = props;
```
shelegdmitriy, line 19: Let's move this to line [4](https://github.com/near/near-discovery-components/pull/520/files#diff-3665dbffa6eb0e091687824986cb9b62bdc720ceb6d9dccf8bdb38d59d67d9b2R4) so all the props will be at the same place at the top

src/NearOrg/Notifications/Notification.jsx
```diff
@@ -322,5 +366,24 @@ return (
...
+      props={{
+        trigger: <i className="ph-bold ph-dots-three" />,
+        items: buildMenu(),
+      }}
+    />
+    <Widget
+      src="${REPL_ACCOUNT}/widget/DIG.Dialog"
+      props={{
+        type: "alert",
+        title: <modalContent />,
```
shelegdmitriy, line 380: I think we can avoid this function and use this instead:
```suggestion
        title: `Do you want to stop receiving push notification for ${type}`,
```

shelegdmitriy (approved):
@erditkurteshiSQA LGTM! I just added a few comments as a cleanup.

src/NearOrg/Notifications/Notification.jsx
erditkurteshiSQA: This has been addressed in the latest commit. Thank you

src/NearOrg/Notifications/NotificationsList.jsx
erditkurteshiSQA: This has been addressed in the latest commit. Thank you

src/NearOrg/Notifications/Notification.jsx
erditkurteshiSQA: This also has been addressed in the latest commit. Thank you

erditkurteshiSQA:
> @erditkurteshiSQA LGTM! I just added a few comments as a cleanup.

Thank you @shelegdmitriy ,
Let's hold on merging until the PR: https://github.com/near/near-discovery/pull/820 is resolved. Once it's addressed, we can continue with the merge.

src/NearOrg/Notifications/Notification.jsx
```diff
@@ -259,10 +262,46 @@ const iconType = {
...
+const buildMenu = () => {
+  return [
+    {
+      name: (
+        <>
+          <i
+            className="ph-bold ph-bell-simple-slash"
+            style={{ color: "#D95C4A" }}
+          />
+          <span style={{ color: "#D95C4A" }}>Block</span>
```
milenakrawczyk, line 276: This should be visible only for users that enabled push notifications. It seems that checking whether a subscription endpoint exists should be enough to determine whether the push notifications have been enabled.

erditkurteshiSQA: I added a condition only for users that have enabled push notifications. Should be fixed in the latest commit. Thank you Milena.

PR #521, opened 2023-11-13
Added handleCloseSearchMenu to close search UI 

erditkurteshiSQA:
Fixes https://github.com/near/near-discovery/issues/807 .

This PR is related with: https://github.com/near/near-discovery/pull/821


calebjacob (changes requested):
See my comment here on the gateway: https://github.com/near/near-discovery/pull/821#discussion_r1391283781

This just means our code would change to this:

```
const handleCloseSearchMenu = () => {
  props.focusChange(false);
}
```

calebjacob (approved):
Thanks for making those changes! :) Looks good.

Issue #447, opened 2023-10-25
DIG.Modal Flashing


calebjacob:
There are a few instances in our BOS UI where `DIG.Modal` is used to render a modal. In most cases, this component will flash multiple times when opening/rendering. It's most noticeable when running the gateway locally with `pnpm run dev`. The `NotificationsAlert` modal is a good example where lots of flashing occurs.

PR #522, opened 2023-11-13
fix: `DIG.Dialog` flashing

shelegdmitriy:
Relates to #447


src/DIG/Dialog.jsx
```diff
@@ -161,6 +161,8 @@ confirmButton = confirmButton ?? (
   />
 );
 
+if (!forwardedProps?.open) return <></>;
```
shelegdmitriy, line 164: I decided to add this condition because without it, this component would render many times on the activity page without even being called

src/DIG/Dialog.jsx
```diff
@@ -38,7 +38,7 @@ const Overlay = styled(`${variant}.Overlay`)`
   position: fixed;
   inset: 0;
   animation: overlayShow 150ms cubic-bezier(0.16, 1, 0.3, 1);
-  z-index: 10000;
+  z-index: 1049;
```
shelegdmitriy, line 41: This change of `z-index` and the next one on line 66 is needed to prevent this modal overlaps:

Знімок екрана 2023-11-13 о 16 35 22

calebjacob (approved):

PR #525, opened 2023-11-13
Code and History Tabs

calebjacob:
Closes: https://github.com/near/near-discovery/issues/814

After talking with @charleslavon, we decided to implement the following on the component details page:

- Shrink the sidebar width a bit to give the code more room
- The `Code` tab now only renders the current code
- The `Preview` tab now only renders the current preview
- The `History` tab shows the commit history and diffs

### DIFF
--- a/src/ComponentHistory.jsx
+++ b/src/ComponentHistory.jsx
@@ -33,16 +33,21 @@ function getTimestampFromBlockHeight(blockHeight) {
 }
 
 const Wrapper = styled.div`
-  display: grid;
-  grid-template-columns: 200px minmax(0, 1fr);
+  display: flex;
+  flex-direction: column;
+  gap: 30px;
+`;
+
+const Layout = styled.div`
+  display: flex;
   gap: 30px;
 
   @media (max-width: 980px) {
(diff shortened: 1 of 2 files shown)


mpeterdev (approved):

src/ComponentHistory.jsx
```diff
@@ -199,125 +214,153 @@ if (!blocksChanges) {
...
+            }}
           />
-        </div>
-      )}
-    </Main>
+        )}
+
+        {state.selectedTab === "history" && (
+          <Widget
+            src={`bozon.near/widget/WidgetHistory.CodeHistory`}
```
shelegdmitriy, line 337: Is this account is fine for testnet too?

shelegdmitriy (approved):
LGTM!

src/ComponentHistory.jsx
calebjacob: Great point - we've never had a `testnet` version of this component and it will continue to cause issues in `testnet`. I opened a separate issue here to track this: https://github.com/near/near-discovery-components/issues/534

PR #528, opened 2023-11-14
Fix Search Typeahead Tabs

calebjacob:
Closes: https://github.com/near/near-discovery/issues/782

- Convert tabs from `<a>` to `<button>` tags
- Fix style issues

### DIFF
--- a/src/Search/Facets.jsx
+++ b/src/Search/Facets.jsx
@@ -111,7 +111,7 @@ const TabsButton = styled.button`
   font-size: 12px;
   padding: 8px 12px;
   position: relative;
-  color: ${(p) => (p.selected ? "#11181C" : "#687076")};
+  color: #687076;
   background: none;
   border: none;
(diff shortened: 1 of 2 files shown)


src/Search/Facets.jsx
```diff
@@ -132,50 +101,41 @@ const Tabs = styled.div`
...
+  all: unset;
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
+  height: 100%;
+  font-weight: 600;
+  font-size: 12px;
+  padding: 8px 12px;
+  position: relative;
+  color: ${(p) => (p.selected ? "#11181C" : "#687076")};
```
shelegdmitriy, line 114: There is no `selected` here, right?

calebjacob: Good catch! Forgot to remove that. Thanks!

shelegdmitriy (approved):

PR #515, opened 2023-11-09
Standalone post and comment views incorporate moderation

gabehamilton:
## Overview
Post and Comment links land on PostPage. It now checks self-moderation and global moderation before fetching and then displaying the data.

### Notes
For readability I settled on doing the three fetches (self-moderation, global moderation, post & comment data) sequentially. With further work this could be refactored to be in parallel, saving 200ms. 

## Testing
All test cases tested locally using bos-loader, links below use the `near` widget location. Self-moderation tests are relative to an account where the post or comment has been 'hidden'.

```[tasklist]
### Test Cases
- [x] Un-moderated post displays: https://near.org/near/widget/PostPage?accountId=cuongdcdev.near&blockHeight=105052885
- [x] Un-moderated comment displays https://near.org/near/widget/PostPage?accountId=cuongdcdev.near&commentBlockHeight=105091974
- [x] Post self-moderation https://near.org/near/widget/PostPage?accountId=jamesbwalc.near&blockHeight=104709179
- [x] Post global moderation https://near.org/near/widget/PostPage?accountId=flatirons.near&blockHeight=104862361
- [x] Comment self-moderation https://near.org/near/widget/PostPage?accountId=robertyan.near&commentBlockHeight=104706273
- [x] Comment global moderation https://near.org/near/widget/PostPage?accountId=masterofcode.near&commentBlockHeight=101133608
- [x] Comment where post is self-moderated https://near.org/near/widget/PostPage?accountId=93aef305b1c24a2df8f2c0de5d1bb71e0b15c0401794cde8a1d13ba2f7f44c32&commentBlockHeight=105100906
- [x] Comment where post is globally moderated https://near.org/near/widget/PostPage?accountId=flatirons.near&commentBlockHeight=105100828
- [x] Invalid link to Post https://near.org/near/widget/PostPage?accountId=cuongdcdev.near&blockHeight=1050528850
- [x] Invalid link to Comment https://near.org/near/widget/PostPage?accountId=cuongdcdev.near&commentBlockHeight=1050919740
- [x] Missing parameters https://near.org/near/widget/PostPage?accountId=cuongdcdev.near
- [x] Moderation of all comments on the post https://near.org/near/widget/PostPage?accountId=marieke.near&blockHeight=100780071
- [x] Comment where account is globally moderated https://near.org/near/widget/PostPage?accountId=govtbot.ndctools.near&commentBlockHeight=100880401
- [x] Check that "" key is handled in matchesModeration and indexer

PR #530, opened 2023-11-15
fix: Rendering post for accounts with no seft moderated data

shelegdmitriy:
Closes: [#828](https://github.com/near/near-discovery/issues/828)


gabehamilton (approved):
Checked all cases from https://github.com/near/near-discovery-components/pull/515 , looks good!

Issue #494, opened 2023-11-06
Inform the users whenever they are not allowed to use Fractal integration feature 


shelegdmitriy:
**Is your feature request related to a problem? Please describe.**
At this moment user can add a couple of verification items at near.org/settings using the idOS SDK. To use idOS there are a couple of required items:

1. Wallet, with which user signed in, should support the `signMessage` method.
2. Wallet, with which user signed in, should be supported by [idos-sdk-js](https://github.com/idos-network/idos-sdk-js).

This means user who logged in with FastAuth will see the error.

**Describe the solution you'd like**
I'd like to see a banner or tooltip message to inform the user about which wallets does support both requirements.

**Additional context**
The list of supported wallets by [idos-sdk](https://github.com/idos-network/idos-sdk-js/blob/main/apps/idos-example-dapp/src/main.js#L73)
To check which wallet supports `signMessage` use this link [signMessage](https://near.org/signMessage)

charleslavon:
Also show the info text without requiring a click, so it appears by default. The recommended wallets text can also appear here.

![Image](https://github.com/near/near-discovery-components/assets/1915116/f507b68f-7194-437c-bb58-ee4838dd18b7)

Issue #493, opened 2023-11-06
Create a tooltip about success connection/creation of fractal account


shelegdmitriy:
**Is your feature request related to a problem? Please describe.**
After visiting a near.org/settings user is able to connect to the idOS network to create credential items. So far there is no clear message about success status of this connection. There is only error shown if something went wrong. 

**Describe the solution you'd like**
After user comes back from the fractal It's good to have a tooltip which informs that connection success

**Additional context**
Task with the design https://github.com/near/near-discovery/issues/647

PR #538, opened 2023-11-21
refactor: Connect to idOS on page load

shelegdmitriy:
This PR introduces a couple of changes:

1. The new tooltip about successful connection to idOS.
2. Added a message to the banner to inform users about preferred wallets to use the idOS feature.
3. Made banner visible by default.
4. Added ability to load idOS data on page load

Relates to: #493 #494

### DIFF
--- a/src/Settings/Identity/Index.jsx
+++ b/src/Settings/Identity/Index.jsx
@@ -1,11 +1,6 @@
 const [showBanner, setShowBanner] = useState(true);
-
-// const modalOpen = fractalModal?.alert ?? false;
-// const fractalModal = Storage.get("fractal-alert") ?? null;
-
-const tooltipState = useMemo(() => props.showTooltip, [props.showTooltip]);
 const [idosData, setIdosData] = useState(null);
-const [showSuccessTooltip, setShowSuccessTooltip] = useState(tooltipState);
+const [showSuccessTooltip, setShowSuccessTooltip] = useState(props.showTooltip);
 
 const Wrapper = styled.div`
   display: flex;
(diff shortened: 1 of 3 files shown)


src/Settings/Identity/Index.jsx
```diff
@@ -1,8 +1,12 @@
-const [showBanner, setShowBanner] = useState(false);
+const [showBanner, setShowBanner] = useState(true);
 
 // const modalOpen = fractalModal?.alert ?? false;
 // const fractalModal = Storage.get("fractal-alert") ?? null;
```
calebjacob, line 4: Does it make sense to get rid of these commented out lines?

src/Settings/Identity/Index.jsx
```diff
@@ -1,8 +1,12 @@
...
 
+const tooltipState = useMemo(() => props.showTooltip, [props.showTooltip]);
```
calebjacob, line 6: This looks like a situation where `useMemo()` isn't providing any value.

src/Settings/Identity/Index.jsx
```diff
@@ -30,6 +34,12 @@ const Icon = styled.i`
 
 const bannerToggle = useCallback(() => setShowBanner(!showBanner), [showBanner]);
 
+useEffect(() => {
+  if (!props.idosConnected && props.connectIdOS) {
+    props.connectIdOS()
+  }
+}, [props.idosConnected]);
```
calebjacob, line 41: Looks like we might have forgot to add `props.connectIdOS` to the dependencies array?

shelegdmitriy: Thanks for catch!

src/Settings/Identity/Index.jsx
shelegdmitriy: Yes. I fixed that logic related to modal

gabehamilton (approved):

PR #547, opened 2023-11-29
Introduce Husky commit hooks

charleslavon:
Adds new dependencies and pre-commit hooks for consistency in our commit messages

affecting only locally dev environments. you can run 'pnpm prepare' to setup the dependencies, then
you can either create commit messages as usual or run cz or pnpm commit for an interactive prompt.

If this looks good I'll add similar configs into near-discovery-components


charleslavon:
@calebjacob do you think it's worth keeping the `printWidth` to a lower value than on discovery in order to have less line wraps in a gateway's view source UI? 
Screen Shot 2023-11-30 at 3 14 35 PM

calebjacob:
@charleslavon Oh that's a great point - probably does make sense to have a smaller line width then. Whatever seems like a good number to you is fine with me.

calebjacob (approved):

PR #563, opened 2023-12-11
feat: Add click directly on TabsButton

erditkurteshiSQA:
Added `onClick` handler directly to the `TabsButton` component to set the selected tab with `setSelectedTab`.
Closes https://github.com/near/near-discovery/issues/693

### DIFF
--- a/src/ActivityPage.jsx
+++ b/src/ActivityPage.jsx
@@ -102,15 +102,27 @@ const TabsButton = styled("Link")`
 return (
   <Wrapper className="container-xl" negativeMargin={selectedTab === "posts"}>
     <Tabs halfMargin={selectedTab === "apps"} noMargin={selectedTab === "posts"}>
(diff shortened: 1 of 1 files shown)


calebjacob (approved):
The `ActivityPage` is by far our heaviest and slowest page and that's probably why the links on this page take a little while to resolve once clicked on less powerful devices. It doesn't hurt to add the `onClick()` logic here to speed things up a little, but I'd be hesitant to add it anywhere else since we should just be relying on the URL state changing.

erditkurteshiSQA:
> The `ActivityPage` is by far our heaviest and slowest page and that's probably why the links on this page take a little while to resolve once clicked on less powerful devices. It doesn't hurt to add the `onClick()` logic here to speed things up a little, but I'd be hesitant to add it anywhere else since we should just be relying on the URL state changing.

You are correct, loading the content will take some time as you mentioned because of the `ActivityPage`.
`onClick()` also will help solve the "lag" with the unselected label becoming bold as described in the issue.

PR #565, opened 2023-12-11
fix: remove explore tab from profile page

jackson-harris-iii:
Closes: #835

### DIFF
--- a/src/ProfilePage.jsx
+++ b/src/ProfilePage.jsx
@@ -224,7 +224,6 @@ return (
           <TabsButton href={`${accountUrl}&tab=followers`} selected={state.selectedTab === "followers"}>
             Followers
           </TabsButton>
-
         </Tabs>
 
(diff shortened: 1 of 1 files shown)


charleslavon (commented):
@jackson-harris-iii I don't feel strongly on whether the Explorer.* components should be deleted, but this should probably also be removed since it's no longer in use here.

charleslavon:
this = https://github.com/near/near-discovery-components/pull/565/files#diff-c6293741ec8490b9af15298970af1135e7827739658c97e3bc7197de49e83f4eR269

jackson-harris-iii:
@charleslavon I went ahead and [removed the additional code](https://github.com/near/near-discovery-components/pull/565/files#diff-c6293741ec8490b9af15298970af1135e7827739658c97e3bc7197de49e83f4eL272) related to displaying the widget.

PR #569, opened 2023-12-13
Refactor NearOrg/Notifications components

shelegdmitriy:
This PR includes a bunch of refactoring to improve performance and code readability.

Closes [#779](https://github.com/near/near-discovery/issues/779)

### DIFF
--- a/src/NearOrg/Notifications/NotificationsList.jsx
+++ b/src/NearOrg/Notifications/NotificationsList.jsx
@@ -18,7 +18,7 @@ const renderItem = (item, i) => {
   return (
     <Widget
       src="${REPL_ACCOUNT}/widget/NearOrg.Notifications.Notification"
-      key={JSON.stringify(item)}
+      key={`${initiator}-${item.blockHeight}-${i}`}
       props={{ initiator, ...item, manageNotification, permission }}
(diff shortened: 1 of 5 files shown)


src/NearOrg/Notifications/NotificationsList.jsx
```diff
@@ -25,236 +18,22 @@ const renderItem = (item, i) => {
   return (
     <Widget
       src="${REPL_ACCOUNT}/widget/NearOrg.Notifications.Notification"
-      key={i}
+      key={JSON.stringify(item)}
```
calebjacob, line 21: Depending on the size of the `item` object, this could potentially create a performance issue. Is there a value or two inside of `item` that we can combine to come up with a unique key instead of stringifying the entire object?

calebjacob (approved):
Looks good! Nice work on the refactor. Left one minor comment.

src/NearOrg/Notifications/NotificationsList.jsx
shelegdmitriy: Yes. There are `accountId`, `blockHeight`. I think we can combine `blockHeight` + `index`

PR #579, opened 2023-12-15
Metadata in components indexer

milenakrawczyk:
This PR adds metadata to the components indexer.

The SQL schema has been changed due to the old version not properly generating.

### DIFF
--- a/indexers/components/components.js
+++ b/indexers/components/components.js
@@ -298,7 +298,7 @@ async function getBlock(block: Block) {
   const lineDiff = new Diff();
   lineDiff.tokenize = function (value) {
     let retLines = [],
-      linesAndNewlines = value.split(new RegExp("(\\n|\r\\n)"));
+    linesAndNewlines = value.split(new RegExp("(\\n|\r\\n)"));
 
     // Ignore the final empty token that occurs if the string ends with a new line
     if (!linesAndNewlines[linesAndNewlines.length - 1]) {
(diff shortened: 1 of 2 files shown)


indexers/components/components.js
```diff
@@ -446,6 +448,13 @@ async function getBlock(block: Block) {
             if (result.removed) linesRemoved += result.count;
           });
 
+          let forkSource = metadata?.fork_of ? metadata?.fork_of.split('@')[0]: null;
+          let forkBlockHeight = metadata?.fork_of ? metadata?.fork_of.split('@')[1]: null;
+          if (forkSource === `${componentAuthorId}/widget/${componentName}`) {
```
milenakrawczyk, line 453: Checking if the fork source is the same as the widget path. If so, the widget was not forked.

indexers/components/components.sql
```diff
@@ -1,14 +1,22 @@
 CREATE TABLE
   "versions" (
-    "id" SERIAL PRIMARY KEY,
-    "block_height" INT NOT NULL,
+    "id" SERIAL,
+    "block_height" BIGINT NOT NULL,
```
gabehamilton, line 4: why the change to bigint?

gabehamilton (approved):

indexers/components/components.sql
milenakrawczyk: This change is optional as it will take a lot of time before we hit the max of INT. However, I figured we may do that now since the schema changes anyway which will result in a new indexer. I may revert it if we stick to INT values only.

PR #585, opened 2023-12-21
NEAR-879 FIX: Updated z-index for profile overlay issue

valmir-haxholli:
Due to Radix Popover having set the z-index to auto, most of the content which have a higher z-index will result in the Profile Overlay showing before content.

This PR will close this issue: [Issue: 879](https://github.com/near/near-discovery/issues/879)

### DIFF
--- a/src/Posts/Post.jsx
+++ b/src/Posts/Post.jsx
@@ -107,7 +107,9 @@ query IndexerQuery {
 
 const Post = styled.div`
   position: relative;
-
+  div[data-radix-popper-content-wrapper] {
+    z-index: 99999 !important;  //to override the inline popover auto index
+  }
   &::before {
     content: "";
     display: block;
--- a/src/Posts/Post.jsx
+++ b/src/Posts/Post.jsx
@@ -107,9 +107,6 @@ query IndexerQuery {
 
 const Post = styled.div`
   position: relative;
-  div[data-radix-popper-content-wrapper] {
-    z-index: 99999 !important;  //to override the inline popover auto index
-  }
   &::before {
     content: "";
     display: block;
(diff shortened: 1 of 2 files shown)


src/Posts/Post.jsx
```diff
@@ -107,7 +107,9 @@ query IndexerQuery {
 
 const Post = styled.div`
   position: relative;
-
+  div[data-radix-popper-content-wrapper] {
+    z-index: 99999 !important;  //to override the inline popover auto index
+  }
```
shelegdmitriy, line 112: Hey @valmir-haxholli. Thanks for your contribution here! This fix works but I would rather modify the `src/AccountProfileOverlay.jsx` file to add `z-index: 1001` for `HoverCardContent` than add this change to the `Post` component.

`z-index: 1001` will be enough because top navbar has `z-index: 1000` and there aren't many places where we use a higher value

shelegdmitriy (changes requested):

src/Posts/Post.jsx
valmir-haxholli: Hey @shelegdmitriy, Thank you for your feedback, based on your suggestion I have updated the `HoverCardContent `component instead, I tested the change and it's seems to work. This change it's more bulletproof then the previous one since the radix attribute might change in the future.

shelegdmitriy: Yeah, and the reason of that is modals. There are couple of places from where we are calling modals from dropdowns. As an example DIG.Dialog or transaction modal. To avoid overlap modals with dropdowns we have to double check those cases. Sure, it's worth to think about how to leverage such cases maybe with introducing some css variables for better control

shelegdmitriy (approved):

PR #587, opened 2023-12-21
fix: fully check data field in indexers

gabehamilton:
Latest indexer code for core social indexers


indexers/social_feed/social_feed.js
```diff
@@ -244,98 +244,115 @@ async function getBlock(block: Block) {
...
+    const actions = block.actions();
+    if (!actions) {
+      console.log("Block has no actions");
+      return;
+    }
+    const contractActions = actions.filter(
+        (action) => action.receiverId === SOCIAL_DB
+    );
+    if (!contractActions) {
+      console.log("Block has no actions");
```
darunrs, line 258: nit: This can be Block has no social actions.

indexers/social_feed/social_feed.js
```diff
@@ -244,98 +244,115 @@ async function getBlock(block: Block) {
...
+                    !functionCall ||
+                    !functionCall.args ||
+                    !functionCall.args.data ||
+                    !Object.keys(functionCall.args.data) ||
+                    !Object.keys(functionCall.args.data)[0]
+                ) {
+                  console.log(
+                      "Set operation did not have arg data in expected format"
+                  );
+                  return;
```
darunrs, line 288: Does return return false or true? We want to skip the ones that don't work right? Or does this return end execution of the block? 

Also, should we also put the function call in the log statement or no? Just depends on if that's useful for debugging or not.

darunrs (approved):

PR #597, opened 2024-01-05
fix: fixed styling bugs in profile editor view

amirsaran3:
Fixes: https://github.com/near/near-discovery/issues/843

The issue here was with using `display: grid` with wrong media breakpoints. It was still rendering the desktop version of main section when it was too small. Since the breakpoint was set to `1024px` and the section was rendered on half of the screen it didn't have room to display correctly.

**Solution:**
Instead of using `display: grid` I used `display: flex` and set it to wrap automatically when there is no more room for the content to display correctly.

Regarding the fourth point in the issue ("'Most recent' dropdown looks buggy and 'Sort by' has incorrect font - bug"):
It looks like the layout is different now and there is no sorting feature anymore.

Before                            |  After
:-------------------------:|:-------------------------:
image | image

### DIFF
--- a/src/ProfilePage.jsx
+++ b/src/ProfilePage.jsx
@@ -39,14 +39,9 @@ const Wrapper = styled.div`
 `;
 
 const Main = styled.div`
-  display: grid;
+  display: flex;
+  flex-wrap: wrap;
   gap: 40px;
-  grid-template-columns: 352px minmax(0, 1fr);
-  align-items: start;
-
-  @media (max-width: 1024px) {
-    grid-template-columns: minmax(0, 1fr);
-  }
 `;
 
 const BackgroundImage = styled.div`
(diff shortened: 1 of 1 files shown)


calebjacob (approved):

gabehamilton:
@amirsaran3 @AmmarHumackicSQA After these changes the regular profile view stacks the tabs beneath the user info instead of having them side by side.

calebjacob:
See comment here: https://github.com/near/near-discovery-components/issues/609#issuecomment-1885224452

amirsaran3:
> @amirsaran3 @AmmarHumackicSQA After these changes the regular profile view stacks the tabs beneath the user info instead of having them side by side.

The tabs will stack when there is no more space in the right column to fit the content. Here is a video of how it looks like when resizeing.
https://github.com/near/near-discovery-components/assets/95852708/0730809a-a6e1-4886-a1ff-2e2a2cac7163

PR #599, opened 2024-01-05
Refactor AccountProfile component

erditkurteshiSQA:
This Pull Request suggests some changes in the `AccountProfile` file to address the issue https://github.com/near/near-discovery/issues/853.

- Added `returnProfileForUser` function.
- Updated `TimeAgo` widget.
- Added prop `profile: returnProfileForUser(profile)` in AccountProfileOverlay widget.

### DIFF
--- a/src/AccountProfile.jsx
+++ b/src/AccountProfile.jsx
@@ -4,15 +4,11 @@ const blockTimestamp = props.blockTimestamp;
 const profileUrl = `/${REPL_ACCOUNT}/widget/ProfilePage?accountId=${accountId}`;
 const verifications = props.verifications;
 const showFlagAccountFeature = props.showFlagAccountFeature ?? false;
-const profile = props.profile;
+const profile = props.profile || Social.get(`${accountId}/profile/**`, "final");
 
 function returnProfileForUser(user) {
-  const rawImage =
-    user.profile_image_1 || user.profile_image_2 || user.profile_image_3;
-  const image =
-    rawImage && rawImage.indexOf("http") === 0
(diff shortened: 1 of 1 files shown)


src/AccountProfile.jsx
```diff
@@ -1,9 +1,36 @@
 const accountId = props.accountId || context.accountId;
-const profile = props.profile || Social.get(`${accountId}/profile/**`, "final");
```
calebjacob, line 2: If we're going to remove the fallback `Social.get(`${accountId}/profile/**`, "final")` logic, we'll need to make sure every use of `<AccountProfile>` is passing in the `profile` prop. I believe there are quite a few areas in our app that don't pass in the `profile` prop to the component and rely on this fallback logic to fetch it based on `accountId`.

calebjacob (changes requested):
Overall this looks cleaner - nice job. See my comment about removing ```Social.get(`${accountId}/profile/**`, "final")```

src/AccountProfile.jsx
shelegdmitriy: I think it's worth to use `Social.get(`${accountId}/profile/**`, "final")` here

shelegdmitriy (changes requested):
Overall LGMT

src/AccountProfile.jsx
erditkurteshiSQA: Thank you for the reviews, I reverted the changes for `Social.get`

calebjacob (approved):

shelegdmitriy (approved):

PR #605, opened 2024-01-08
Enable sending notifications for interactions on comment

amirsaran3:
Fixes: https://github.com/near/near-discovery/issues/668

There is already logic to handle notification sending when liking comment. 

The issue was with `notifyAccountId` being set to `undefined` in `src/Comments/Comment.jsx`. When checking whether to send notification `notifyAccountId` needs to be truthy. Since it was set to `undefined` and never modified after that it would not send the notification.

I also added a check to prevent sending notifications to yourself when liking own comment.

### DIFF
--- a/src/Comments/Comment.jsx
+++ b/src/Comments/Comment.jsx
@@ -11,16 +11,8 @@ State.init({
   showToast: false,
   flaggedMessage: { header: "", detail: "" },
   content: JSON.parse(props.content) ?? undefined,
-  notifyAccountId: undefined,
+  notifyAccountId: accountId,
 });
-
-const extractNotifyAccountId = (parentItem) => {
-  if (!parentItem || parentItem.type !== "social" || !parentItem.path) {
-    return undefined;
-  }
-  const accountId = parentItem.path.split("/")[0];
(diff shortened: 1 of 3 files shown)


src/Comments/Comment.jsx
```diff
@@ -275,7 +267,7 @@ return (
                   src="${REPL_ACCOUNT}/widget/Comments.Compose"
                   props={{
                     initialText: `@${accountId}, `,
-                    notifyAccountId: extractNotifyAccountId(state.content.item),
+                    notifyAccountId: accountId,
```
calebjacob, line 270: I'm worried this change might break the notification that's sent to the post owner when someone creates a new comment. We need to make sure that:

1. New comments notify the original post author that someone left a comment on their post.
2. A comment that is liked sends a notification to the comment author.

calebjacob (changes requested):

src/Comments/Comment.jsx
amirsaran3: I've reverted the changes regarding the `extractNotifyAccountId` function.

calebjacob (approved):

shelegdmitriy (approved):

PR #607, opened 2024-01-09
feat: idOS design enhancement

shelegdmitriy:
This PR introduces [design](https://www.figma.com/file/tub847jdGUsekBXok7gAjs/idOS-settings-redesign?node-id=7%3A8&mode=dev) enhancements.
A couple of things I made different but I can record a loom video describing all the changes I've made.

Closes: [#885](https://github.com/near/near-discovery/issues/885)
Related to https://github.com/near/near-discovery/pull/913


charleslavon:
@shelegdmitriy a loom would be helpful. 

What is the expected UX when landing on Settings.Verifications.Index for a user who has already verified humanness?  I'm setting the UI show the page as if I have not yet setup idos, the only option it gives is to start the setup flow again. 

Screen Shot 2024-01-11 at 10 19 36 AM

charleslavon:
Ah, looks like [this PR ](https://github.com/near/near-discovery/pull/913) must be merged before this UX can be tested.

shelegdmitriy:
> Ah, looks like [this PR ](https://github.com/near/near-discovery/pull/913) must be merged before this UX can be tested.

That's right

charleslavon:
@shelegdmitriy what's the expected UX here? I'm connected via mynearwallet, I'm already verified, but I get the UX as if I have never set this up. 
Screen Shot 2024-01-12 at 1 00 24 PM

shelegdmitriy:
> @shelegdmitriy what's the expected UX here? I'm connected via mynearwallet, I'm already verified, but I get the UX as if I have never set this up.  This is with bos-loader serving the components from this PR on beta.near.org which has your latest PR. 
> 
> 
> 
> Screen Shot 2024-01-12 at 1 00 24 PM
> 
> 
> 
> 

Could you try to clear your local storage?

charleslavon:
@shelegdmitriy I experience the same when connecting on an incognito window.

shelegdmitriy:
> @shelegdmitriy I experience the same when connecting on an incognito window.

That's very interesting because it seems I found a couple of gaps in my code. Let me double check it and I'll ping you to test on your end. Sorry for that

src/Settings/Identity/Banner.jsx
```diff
@@ -66,30 +64,7 @@ return (
...
-            HERE Wallet,
-          </TextLink>{" "}
-          <TextLink href="https://wallet.nightly.app" target="_blank">
-            Nightly Wallet,
-          </TextLink>{" "}
-          <TextLink href="https://app.mynearwallet.com" target="_blank">
-            MyNearWallet
-          </TextLink>{" "}
-        </Text>
+        <Text>To begin using and displaying identity data in your settings, completed the onboarding steps below.</Text>
```
charleslavon, line 67: @shelegdmitriy I think this should be removed, since it doesn't apply to users who have already linked a profile and we have the two images that direct the user to the next steps.

shelegdmitriy: Do you mean to remove this totally or leave for users who doesn't have any verifications yet?

src/Settings/Identity/Onboarding/Cards.jsx
```diff
@@ -0,0 +1,181 @@
...
+  padding: 0 24px;
+`;
+
+return (
+  <Wrapper>
+    <Card background="var(--white)">
+      <Step background="var(--sand12)" color="var(--white)">
+        1
+      </Step>
+      <Title>Log into idOS compatible wallet</Title>
```
charleslavon, line 113: @shelegdmitriy  Once this is merged and we can test the end-to-end flow, let's consider not displaying this for users who are already connected to an iDOS compatible wallet.

src/Settings/Identity/Onboarding/Cards.jsx
```diff
@@ -0,0 +1,181 @@
...
+          operating system (idOS). To view your data, sign up or sign in.
+        </Text>
+        <Text color="var(--sand10)">
+          Learn more about idOS
+          <TextLink href={idOSLearnLink} target="_blank">
+            here
+          </TextLink>
+        </Text>
+      </TextWrapper>
+      <Widget
```
charleslavon, line 161: @shelegdmitriy Do we know enough about the user's iDOS state to be able to reduce this to only one `Connect` button?

shelegdmitriy: Actually we don't need two buttons here. I asked similar question in figma about this.

charleslavon (approved):

PR #618, opened 2024-01-19
Moderators can specify the reason for moderation

gabehamilton:
When moderating content or an account, moderators are now presented with a dialog to choose the reason. This same dialog will be also used by users when reporting (future PR).


src/DIG/Dialog.jsx
```diff
@@ -202,7 +202,7 @@ return (
         )}
       </ActionWrapper>
       {(type === "dialog" || enableCloseButton) && (
-        <CloseButtonWrapper>
+        <CloseButtonWrapper onClick={onCancel}>
```
shelegdmitriy, line 205:
```suggestion
        <CloseButtonWrapper>
```

No need to pass another prop here. Just use [onOpenChange](https://www.radix-ui.com/primitives/docs/components/alert-dialog#root) in `ReasonDialog` component

src/Moderation/NeedsModeration.jsx
```diff
@@ -289,19 +292,16 @@ const renderItem = (item) => {
...
                 tooltip: blockAccountHelperText + (isModerator ? "" : "\n" + disabledMessage),
                 variant: "destructive",
                 fill: "outline",
-                data: {
-                  index: {
-                    moderate: moderationDataFormat(accountId),
-                  },
-                },
+                type: pathToType("account"),
+                buildData: moderationDataFormat.bind(null, accountId, null, null),
```
shelegdmitriy, line 304:
```suggestion
                buildData: moderationDataFormat(null, accountId, null, null),
```

I believe there is no need to use `.bind`

src/Moderation/ReasonDialog.jsx
```diff
@@ -0,0 +1,150 @@
+const type = props.type;
+const open = props.open;
+const closeModal = props.closeModal;
+const submitClick = props.submitClick;
```
shelegdmitriy, line 4:
```suggestion
let {
  type,
  open,
  closeModal,
  submitClick
} = props;
```

src/Moderation/ReasonDialog.jsx
```diff
@@ -0,0 +1,150 @@
...
+          <h4>Report {type}</h4>
+        </div>
+        <div className="d-flex flex-row gap-2">
+          <span>Tell us more about why this {type} is problematic.</span>
+        </div>
+      </div>
+      <div className="d-flex flex-column gap-3">
+        {Object.keys(reasons).map((reason) => {
+          return (
+            <div className="d-flex flex-row gap-2">
```
shelegdmitriy, line 69: Please provide an unique `key` for the element:

```suggestion
            <div className="d-flex flex-row gap-2" key={reason.id // or what is the unique value here}>
```

src/Moderation/ReasonDialog.jsx
```diff
@@ -0,0 +1,150 @@
...
+            </div>
+          );
+        })}
+      </div>
+      <div className="d-flex flex-row gap-2">
+        <InfoText>
+          <i className="ph-bold ph-info" /> Unsure? Review our community{" "}
+          <a href="https://near.org/terms" target="_blank">
+            content policies.
+          </a>
```
shelegdmitriy, line 94:
```suggestion
          <Link href="https://near.org/terms" target="_blank">
            content policies.
          </Link>
```

src/Moderation/ReasonDialog.jsx
```diff
@@ -0,0 +1,150 @@
...
+  font-size: 1em;
+`;
+
+return (
+  <Widget
+    src="${REPL_ACCOUNT}/widget/DIG.Dialog"
+    props={{
+      type: "dialog",
+      description: buildContent(type),
+      onCancel: closeModal,
```
shelegdmitriy, line 144:
```suggestion
      onOpenChange: closeModal,
```

shelegdmitriy (changes requested):

src/Moderation/NeedsModeration.jsx
gabehamilton: This function gets passed in to the dialog and called there with the reason.

shelegdmitriy: Your method is written as an arrow function that's why no need to use `bind` there. Try that out 🙂

src/Moderation/ReasonDialog.jsx
gabehamilton: React isn't complaining about this for me since it's not a React component. Why do we need a key?

calebjacob (approved):
Looks good @gabehamilton - let's just make sure we address the `key` comment above before merging.

PR #619, opened 2024-01-19
feat: updated components.sql to capture stars/forks and additional info

Kevin101Zhang:
created a new table "info" to capture forks, stars, and additional metadata for components

### DIFF
--- a/indexers/components/components.sql
+++ b/indexers/components/components.sql
@@ -1,7 +1,7 @@
 CREATE TABLE
   "versions" (
     "id" SERIAL PRIMARY KEY,
-    "block_height" INT NOT NULL,
+    "block_height" BIGINT NOT NULL,
     "block_timestamp_ms" BIGINT NOT NULL,
     "code" VARCHAR NOT NULL,
     "component_author_id" VARCHAR NOT NULL,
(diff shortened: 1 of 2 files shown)


gabehamilton (approved):

indexers/components/components.sql
```diff
@@ -1,22 +1,33 @@
 CREATE TABLE
   "versions" (
-    "id" SERIAL,
-    "block_height" BIGINT NOT NULL,
+    "id" SERIAL PRIMARY KEY,
+    "block_height" INT NOT NULL,
```
calebjacob, line 4: I remember us running into issues using an `INT` instead of `BIGINT` here for `block_height`. Are we sure `INT` is within range? @gabehamilton

Kevin101Zhang: "The maximum value for INT is 2,147,483,647 (2^31-1) while the maximum value for BIGINT is 9,223,372,036,854,775,807 (2^63-1)."

I dont understand the issue here. To my knowledge - The bigint type should only be used if the range of the integer type is insufficient, because the latter is definitely faster. We're saving 4bytes per entry with INT with our latest block height at 111,056,090 no where near the cap with over 2 billion leftover currently.

On the contrary I believe it could be argued that storage is cheap. The fire drill when you run out of integers is not cheap.

pkudinov: It might appear that it will take > 60 years to reach MAX_INT block height (assuming 1 block per second), but in this case I would say it's best to use types that are as close to definitions as possible. The block height type is defined here: https://github.com/near/nearcore/blob/55f5586e89708f0b23f6a57aae58eaef270e77af/chain/indexer/src/lib.rs#L61 and I think it's best to use the same type here. Which is `BIGINT`.

PR #631, opened 2024-01-31
fix: fixed notification content description

erditkurteshiSQA:
This pull request resolves the first point raised in issue https://github.com/near/near-discovery/issues/909 which fixes the content description not matching the action.

- Introduction of the `notificationValue` property for `getNotificationContent` function.
- Added `commentAuthorAccountId` to get the specific accountId
- Addition of `contentBlockHeight` to accurately capture the blockHeight for notifications of type `like`.
- Refactored `contentPath` and `contentDescription`.


shelegdmitriy (approved):
Seems like I missed this part on refactoring this things. Good catch @erditkurteshiSQA !

PR #634, opened 2024-02-03
feat: use accounts indexer for posts feed

### LINKED ISSUE #427: Feeds should pull in profile fields via graphql relationships

shelegdmitriy:
Relates to [#855](https://github.com/near/near-discovery/issues/855)
Closes #427 

- [x] add `relationship.json` file

### DIFF
--- a/src/AccountProfile.jsx
+++ b/src/AccountProfile.jsx
@@ -4,29 +4,42 @@ const blockTimestamp = props.blockTimestamp;
 const profileUrl = `/${REPL_ACCOUNT}/widget/ProfilePage?accountId=${accountId}`;
 const verifications = props.verifications;
 const showFlagAccountFeature = props.showFlagAccountFeature ?? false;
-const profile = props.profile || Social.get(`${accountId}/profile/**`, "final");
+console.log(accountId, "AccountProfile profile: ", props.profile);
+// const profile = props.profile || Social.get(`${accountId}/profile/**`, "final");
+const profile = props.profile;
+const chainProfile = Social.get(`${accountId}/profile/**`, "final");
+console.log(accountId, "AccountProfile chainProfile: ", chainProfile);
 
-function returnProfileForUser(user) {
-  const rawImage = user.profile_image_1 || user.profile_image_2 || user.profile_image_3;
-  const image = rawImage && rawImage.indexOf("http") === 0 ? { url: rawImage } : { ipfs_cid: rawImage };
-  const name = user.profile_name ?? "";
-  let tags = null;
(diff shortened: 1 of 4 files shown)


gabehamilton:
Ultimately the AccountProfile widget receives a `profile` property. Maybe the Feed code can be cleaned up so the account relationship (which could be renamed profile) gets passed all the way down and post/item doesn't have profile_image, etc fields?

src/AccountProfile.jsx
```diff
@@ -1,32 +1,11 @@
...
-    };
-  }
+accountId = accountId || context.accountId;
+showFlagAccountFeature = showFlagAccountFeature ?? false;
+profile = profile || Social.get(`${accountId}/profile/**`, "final");
+
+const profileUrl = `/${REPL_ACCOUNT}/widget/ProfilePage?accountId=${accountId}`;
 
-  return null;
-}
```
shelegdmitriy, line 29: I dropped this method because from now QueryAPI returns `image` and `tags` in format we need. Also I noticed that `Social.get(`${accountId}/profile/**`, "final")` doesn't have a `profile_image_1` or `profile_image_2` and returns `image` prop (as an example) in format:
```
image: {
  ipfs_cid: "asdasdasdasdasdasdasd"
}
or
image: {
  nft: {
    contractId: "asasdasda",
    tokenId: "asdasdasdasd"
  }
}
```
and I don't see any reason to use this method here

gabehamilton (approved):
Nice!

shelegdmitriy:
@gabehamilton may I ask you to verify that `relationship.json` file correctly updated

gabehamilton:
The relationship looks good
Screenshot 2024-02-12 at 10 18 07 AM

shelegdmitriy:
@calebjacob I'd like you to take a look at this changes

calebjacob (approved):
Great job! Glad you were able to clean up a lot of this. :)

PR #642, opened 2024-02-09
feat: use `accounts` indexer for notifications

shelegdmitriy:
Relates to: [#855](https://github.com/near/near-discovery/issues/855)

TODO list:

- [x] add limits to graphql queries;
- [x] add fallback logic similar to `feeds`;
- [x] add `relationship.json`

Problem:

- [x] There is no `post` param for devhub posts. Maybe need to check logic of `notifications` indexer and how it collects devhub posts. That's why notifications from devhub doesn't have a preview content using graphql query


shelegdmitriy:
Hey @gabehamilton! I'd like to hear your opinion about how we'll solve the problem with missing fields for devhub posts. I'll check what else are missing and maybe it's worth to update the `notifications` indexer

src/QueryApi/utils.jsx
```diff
@@ -0,0 +1,20 @@
...
+      query: operationsDoc,
+      variables: variables,
+      operationName: operationName,
+    }),
+  });
+}
+
+return {
+  fetchGraphQL,
+};
```
shelegdmitriy, line 20: I think this file is worth to move to another PR

charleslavon:
> Hey @gabehamilton! I'd like to hear your opinion about how we'll solve the problem with missing fields for devhub posts. I'll check what else are missing and maybe it's worth to update the `notifications` indexer

Is what's needed here the item.blockHeight? There is a similar need described [here](https://github.com/near/near-discovery/issues/658)

shelegdmitriy:
> > Hey @gabehamilton! I'd like to hear your opinion about how we'll solve the problem with missing fields for devhub posts. I'll check what else are missing and maybe it's worth to update the `notifications` indexer
> 
> Is what's needed here the item.blockHeight? There is a similar need described [here](https://github.com/near/near-discovery/issues/658)

No. In general this was about the missing `post` prop but it's already added to the notifications indexer and updated. I'll update this PR soon

shelegdmitriy:
After checking I realized I need to do one more update of `notifications` indexer with the needed field. Thanks @charleslavon!

gabehamilton (approved):

src/NearOrg/Notifications/NotificationsMiddleware.jsx
```diff
@@ -0,0 +1,127 @@
...
+  query IndexerQuery {
+    data: dataplatform_near_notifications_notifications( limit: 1, order_by: { block_height: desc }) {
+      blockHeight
+    }
+  }
+`;
+
+const notificationsQuery = `query NotificationsQuery($offset: Int, $limit: Int) {
+  data: dataplatform_near_notifications_notifications(
+    distinct_on: blockHeight,
```
shelegdmitriy, line 15: Hey @gabehamilton! According to our private discussion, I tried to use two fields here like so:
```suggestion
    distinct_on: [blockHeight, initiatedBy],
```
but also I should update `order_by` like so
```
    order_by: {blockHeight: desc, initiatedBy: asc},
```

but after this steps I received a list of data which is ordered in a very strange way. Am I missed something?

shelegdmitriy: Changing `initiatedBy: asc` => `initiatedBy: desc` doesn't help

gabehamilton: I see the same behavior in the GraphiQL editor. It looks like the `initiatedBy` order is taking precedence, regardless of order in the query.

gabehamilton: Ah, order by should be `order_by: [{blockHeight: desc}, {initiatedBy:asc}]`

PR #644, opened 2024-02-11
feat: an EntityList with example usage

gabehamilton:


calebjacob (approved):
Looks good! Should help cut down on some repeated boilerplate.

PR #648, opened 2024-02-14
feat: create a new component for `papers` page

shelegdmitriy:
Relates to https://github.com/near/near-discovery/pull/971

### DIFF
--- a/src/NearOrg/PapersPage.jsx
+++ b/src/NearOrg/PapersPage.jsx
@@ -1,0 +1,225 @@
+const papersDocs = [
+  {
+    name: "The NEAR White Paper",
+    description:
+      "This document describes the approach NEAR takes to designing and implementing the core technology of its system.",
+    icon: "ph-file-doc",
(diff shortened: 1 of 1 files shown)


src/NearOrg/PapersPage.jsx
```diff
@@ -0,0 +1,225 @@
+const papersDocs = [
+  {
+    name: "The NEAR White Paper",
+    description:
+      "This document describes the approach NEAR takes to designing and implementing the core technology of its system.",
+    icon: "ph-file-doc",
+    url: "https://drive.google.com/file/d/1qxlYhQSvh_rpdI_nQMQF9C_QSWIstY2H/view?usp=sharing",
```
shelegdmitriy, line 7: @charleslavon  This is a link to the Google Drive where I uploaded these documents. Let me know if this is a bad idea

charleslavon: let's use our gateway's public folder to serve these PDFs as regular static assets.

PR #656, opened 2024-02-21
feat: simplifying the action buttons on the component details card

charleslavon:
as per [#995](https://github.com/near/near-discovery/issues/955)

### DIFF
--- a/src/ComponentSummary.jsx
+++ b/src/ComponentSummary.jsx
@@ -33,10 +33,12 @@ const primaryActions = {
   open: {
     display: "Open",
     url: appUrl,
+    fork: context.accountId === accountId ? false : true,
   },
   viewDetails: {
     display: "View Source",
     url: detailsUrl,
+    fork: false,
   },
 };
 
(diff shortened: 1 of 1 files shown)


src/ComponentSummary.jsx
```diff
@@ -245,14 +245,10 @@ return (
...
-        {context.accountId === accountId ? (
+        {context.accountId === accountId && (
           <>
             <i className="bi bi-pencil-fill"></i> Edit
           </>
-        ) : (
-          <>
-            <i className="bi bi-git"></i> Fork
-          </>
         )}
```
shelegdmitriy, line 252: After this change there is a button without text on it:
 
Знімок екрана 2024-02-22 о 14 35 27

What is the expected behavior here?

shelegdmitriy (changes requested):

src/ComponentSummary.jsx
charleslavon: @shelegdmitriy thanks for finding this.  I've fixed this with the latest commit. I would've preferred to refactor this component into separate components for the distinct use cases of this component details page versus the Component Card in the develop navigation menu, but instead of that I've changed the card logic to ensure that the Component Card in the develop menu shows the Edit button only when viewing a component that you own, and otherwise shows only one button View Source.

calebjacob: @charleslavon I totally agree with eventually breaking out a new/separate component for the `Current Component` in the main nav dropdown. These two use cases have diverged quite a bit over time and will probably continue to diverge further. Whenever this refactor happens, take a look at `navigation/CurrentComponent.tsx` in our gateway. There's some super hacky CSS in that component that would be great to rip out.

shelegdmitriy (approved):

PR #664, opened 2024-02-23
adds a custom cookie prompt to replace iubenda

charleslavon:
closes https://github.com/near/near-discovery/issues/874


src/DIG/Dialog.jsx
```diff
@@ -22,7 +22,7 @@ let {
 type = type ?? "dialog";
 enableCloseButton = enableCloseButton ?? false;
 
-if (["alert", "dialog"].indexOf(type) < 0) {
+if (["alert", "dialog", "custom"].indexOf(type) < 0) {
```
calebjacob, line 25: Would it work to use the existing `enableCloseButton` (setting it false) instead of needing to add this `custom` type?

calebjacob (approved):
Looks good! Just left one small comment. Feel free to merge whenever.

src/DIG/Dialog.jsx
charleslavon: It doesn't work. Due to all 'dialog' types getting a close button

calebjacob: Oh I see - you're right. That seems like something we should fix. We could still remove the `custom` type and then just update `enableCloseButton` like so:

```
enableCloseButton = enableCloseButton ?? type === 'dialog';
```

PR #686, opened 2024-03-05
chore: add placeholders to events page

shelegdmitriy:
There might be a case when there won't be any upcoming events so this PR introduces a couple of placeholders to inform users about that.

### DIFF
--- a/src/Events/Index.jsx
+++ b/src/Events/Index.jsx
@@ -204,7 +204,7 @@ const IconCircle = styled.div`
   }
 `;
 
-const featuredEvent = [...eventsList].sort(
+const featuredEvent = eventsList.lenght > 0 && [...eventsList].sort(
   (a, b) => Math.abs(new Date(a.event.start_at) - new Date()) - Math.abs(new Date(b.event.start_at) - new Date()),
 )[0].event;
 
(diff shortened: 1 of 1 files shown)


src/Events/Index.jsx
```diff
@@ -204,7 +204,7 @@ const IconCircle = styled.div`
   }
 `;
 
-const featuredEvent = [...eventsList].sort(
+const featuredEvent = eventsList.lenght > 0 && [...eventsList].sort(
```
charleslavon, line 207: `.length`

src/Events/Index.jsx
```diff
@@ -273,30 +276,52 @@ return (
...
-                  location,
-                  target: "_blank",
-                  rel: "noopener noreferrer",
-                  loading: !dataLoaded,
-                }}
-              />
-            );
-          })}
-        </Grid>
+        {eventsList.lenght > 0 ? (
```
charleslavon, line 279: `.length`

shelegdmitriy: That's why it would be good to have some linter 🙂

charleslavon (approved):

PR #692, opened 2024-03-10
Added devhub proposal notification

Megha-Dev-19:


Megha-Dev-19:
@frol I have added `devhub/like` and `devhub/reply` to UI, so that the user is notified correctly about proposal (instead of devhub.near account), let me know if there something else you expect here.

src/NearOrg/Notifications/utils.jsx
```diff
@@ -105,6 +114,17 @@ function createNotificationLink(notificationType, notificationValue, authorAccou
...
+    case "devhub/mention":
+    case "devhub/edit":
+    case "devhub/reply":
+    case "devhub/like":
+      return href({
+        widgetSrc: "devhub.near/widget/app",
+        params: {
+          page: "proposal",
+          id: proposal,
+        },
```
frol, line 126: @Megha-Dev-19 I think you also need to pass the [`notifier` to the parameters](https://github.com/NEAR-DevHub/neardevhub-contract/commit/46ca5c3f2b469e1a56b5773ae6224928b51a1649)

Megha-Dev-19: Oh I didn't push the commit, just did, update the profile as well

frol (approved):

frol:
@charleslavon Can you help with the review or assigning someone to review and merge/release it?

shelegdmitriy:
LGTM! Seems like there are two new fields which comes from `value`: `proposal` and `notifier`. Both of them has `string` type, right? I'm asking this because we [released](https://github.com/near/near-discovery-components/pull/642) usage of `notification` indexer for addition page. And seems this two fields are missed there and I'm happy to help with this! Also, we'll need to update our service worker. All I need from you is to confirm that the types of new fields are correct so I could create a task to update `notifications` indexer.

cc @frol @Megha-Dev-19

shelegdmitriy (approved):

Megha-Dev-19:
@shelegdmitriy hello, you're right about notifier, but for proposal it will be an integer.

shelegdmitriy:
> proposal

Thanks @Megha-Dev-19! I created an issue for this. May I ask you to provide me some accountId which can this new notification type for testing please?

Megha-Dev-19:
@shelegdmitriy yes you can use `devhub.near`.

shelegdmitriy:
Thank you @Megha-Dev-19 for your contribution!

PR #699, opened 2024-03-13
Make changes to handle new signup flow + comments

jackson-harris-iii:


src/LikeButton.jsx
```diff
@@ -108,7 +108,7 @@ const likeClick = () => {
 const title = hasLike ? "Unlike" : "Like";
 
 return (
-  <LikeButton disabled={state.loading || dataLoading || !context.accountId} title={title} onClick={likeClick}>
```
charleslavon, line 111: is this still necessary?

charleslavon (approved):
@jackson-harris-iii the logged out experiencing of viewing a blog and being prompted to login in order to like a post works very smooth. Nice.  

One minor comment on src/LikeButton. Otherwise this LGTM

src/LikeButton.jsx
jackson-harris-iii: I think so, otherwise a like can be added by a user that isn't logged in.

PR #704, opened 2024-03-15
feat: add promote menu option, write addToBlog to socialDb

jackson-harris-iii:


src/Posts/Menu.jsx
```diff
@@ -77,6 +77,44 @@ const moderate = (account, block, messageKey, action) => {
...
+  State.update({
+    loading: true,
+  });
+
+  const data = {
+    index: {
+      promote: JSON.stringify({
+        key: context.accountId,
+        value: {
+          operation: "addToBlog",
```
gabehamilton, line 101: After thinking about it more, maybe this could be `type: "blog"` the operations are: add, delete. `add` can be the default.

gabehamilton, line 101:
```suggestion
          type: "blog",
```
After thinking about it more, maybe this could be type: "blog" the operations are: add, delete. add can be the default.

src/Posts/Menu.jsx
```diff
@@ -94,7 +132,14 @@ const buildMenu = (accountId, blockHeight) => {
     },
   ];
 
+  const promoteToBlogSubmenu = [];
+
   if (blockHeight) {
+    promoteToBlogSubmenu.unshift({
```
gabehamilton, line 138: You might also need to check the item type (path?) to see that it is /post/main since this menu also applies to comments (and other content in the future).

gabehamilton (approved):
Assuming comments get addressed

jackson-harris-iii:
> Assuming comments get addressed

So I completed the suggestions but held this PR until the indexer work is done. However it is more of a draft at this time because it is using my personal indexer and modifying the data_moderation indexer is required along with updating the components in this PR to use the indexer that we will modify

indexers/social_feed/social_feed.sql
```diff
@@ -1,55 +1,66 @@
...
+    "post_id" SERIAL NOT NULL,
+    "account_id" VARCHAR NOT NULL,
+    "block_height" DECIMAL(58, 0),
+    "block_timestamp" DECIMAL(20, 0) NOT NULL,
+    "receipt_id" VARCHAR NOT NULL,
+    CONSTRAINT "post_likes_pkey" PRIMARY KEY ("post_id", "account_id")
+  );
+  
+CREATE TABLE
+  "promote" (
```
charleslavon, line 37: would be more explicit if titled `promoted_post`

indexers/social_feed/social_feed.js
```diff
@@ -13,17 +13,19 @@ import { Block } from "@near-lake/primitives";
  * @param {block} Block - A Near Protocol Block
  */
 async function getBlock(block: Block) {
+  console.log(block.actions());
```
charleslavon, line 16: are all the instances of console.logs left intentionally to support debugging?

indexers/social_feed/social_feed.js
```diff
@@ -87,18 +89,18 @@ async function getBlock(block: Block) {
...
           const currentTimestamp = Date.now();
           await context.db.Posts.update(
-              { id: post.id },
-              { last_comment_timestamp: currentTimestamp }
+            { id: post.id },
+            { last_comment_timestamp: currentTimestamp }
           );
           console.log(`Comment by ${accountId} has been added to the database`);
         } catch (e) {
           console.log(
```
charleslavon, line 97: please log these as error `console.error`

charleslavon: ah these are unchanged, just formatting differences

indexers/social_feed/social_feed.js
```diff
@@ -241,6 +243,82 @@ async function getBlock(block: Block) {
...
+          post_id: post.id,
+        };
+
+        // Call GraphQL mutation to insert a new promotion
+        await context.db.Promote.insert(promotionData);
+
+        console.log(`Promotion by ${accountId} has been added to the database`);
+      }
+    } catch (e) {
+      console.log("Error handling add promotion", JSON.stringify(e));
```
charleslavon, line 318: noted that this file has an existing pattern of logging errors as `log` instead of `error`. @gabehamilton where do these logs output? Should we improve the logging by writing errors as `error`?

gabehamilton:
Seems like there are a lot of whitespace changes. Is this in the formatting that the QueryApi component uses?

indexers/social_feed/social_feed.sql
gabehamilton: I think we can promote lots of things, though you are right that we're making use of post_id for everything here.
`promotion_type` is blog for current data.

indexers/social_feed/social_feed.js
```diff
@@ -241,6 +243,82 @@ async function getBlock(block: Block) {
...
   }
 
+    async function handlePromotion(
+    accountId,
+    blockHeight,
+    blockTimestamp,
+    receiptId,
+    promoteString
+  ) {
+    const promotion = JSON.parse(promoteString);
```
gabehamilton, line 253: This function could use a try/catch so that it's the same as the other handlers. Specifically, someone will eventually write some invalid data to their profile and break that JSON.parse.

jackson-harris-iii:
> Seems like there are a lot of whitespace changes. Is this in the formatting that the QueryApi component uses?

my guess is the prettier wasn't ran last time this file was updated.

indexers/social_feed/social_feed.js
jackson-harris-iii: will do. I refactored the function into two functions handlePromotion and handleAddPromotion and in doing so brought the try catch into the latter without updating the former.

gabehamilton:
> > Seems like there are a lot of whitespace changes. Is this in the formatting that the QueryApi component uses?
> 
> my guess is the prettier wasn't ran last time this file was updated.
Ok great. Looks like your changes match the QueryApi.app formatter. Thanks for fixing that.

jackson-harris-iii:
Last review request on this one. Went ahead and completed the details for the [issue](https://github.com/near/near-discovery/issues/1086) related to adding the blog to the user profile page (this commit allows for a return path back to a user profile within the blog itself)

gabehamilton (approved):

PR #720, opened 2024-03-21
feat: implement new homepage for logged in users

shelegdmitriy:
Closes https://github.com/near/near-discovery/issues/1071


shelegdmitriy:
The reason why I named this new page like `HomePageV2` is because we already have a several `HomePage` components. I'd like to ask @calebjacob @charleslavon how do we want to store new pages like this?

charleslavon:
Hey @shelegdmitriy thanks for asking. Since this layout is specific to the logged-in state, let's move away from the `NearOrg.` prefix and use a new one, such as `Gateway.Home`

calebjacob:
Nice work @shelegdmitriy! I agree with Charles on the naming.

I just merged all my layout change PR's and I wonder if it'd be helpful to use the new container class on this page too:

```
<Wrapper className="gateway-page-container">
```

Then your wrapper styles could simplify to this:

```
const Wrapper = styled.div`
  --section-gap: 20px;
  --text-hero: 500 56px/1 "FK Grotesk", "Mona Sans", sans-serif;
  gap: var(--section-gap);
  display: flex;
  flex-direction: column;
  max-width: 960px;
`;
```

You'd still need to set `max-width: 960px;` since that's different from the container's default `max-width: 1320px;`.

calebjacob (approved):

PR #724, opened 2024-03-27
fix: broken devhub notifications

### LINKED ISSUE #717: Devhub Notifications are broken
- I recently added [PR](https://github.com/near/near-discovery-components/pull/692) to add support for different devhub notifications types, starting with `devhub/`, but it doesn't work as expected. 
I have a created a [video](https://wetransfer.com/downloads/81d464364214382356aa934db963d47a20240321143816/217111?utm_campaign=TRN_TDL_05&utm_source=sendgrid&utm_medium=email&trk=TRN_TDL_05) to explain the issue, the code also breaks for previously added `devgovgigs/` notifcations (on main branch) which worked fine previously.

shelegdmitriy:
This PR introduces an extra check for notification `value` param which needed to show notification properly. 

### Why do we need this check?

Well, we query for the notification data using queryApi. This means some params could be missed because of some new notification type or because the data itself has been created long time ago and indexer didn't collected them. That's why some notifications could be hidden for user or some parts of notification might not exist (as notification preview for example or entire `custom` notification).

### Solution

Each notification type has required amount of params which needed to create a human readable message. So, I've added a method to check whatever a notification value (from QueryAPI) contain all the needed params. This check method depends on type of notification (this param comes from QueryAPI all the time. I believe so 🙂) If the notification does not have a certain value necessary for its display, we make a `Social.get()` request (the same as we have before QueryAPI implementation).

Closes #717

### DIFF
--- a/src/NearOrg/Notifications/Notification.jsx
+++ b/src/NearOrg/Notifications/Notification.jsx
@@ -1,5 +1,20 @@
-const { createNotificationMessage, getNotificationContent, createNotificationLink, getNotificationIconClassName } =
-  VM.require("${REPL_ACCOUNT}/widget/NearOrg.Notifications.utils") || (() => {});
+const {
+  createNotificationMessage,
+  getNotificationContent,
+  createNotificationLink,
+  getNotificationIconClassName,
+  checkNotificationValueAvailability,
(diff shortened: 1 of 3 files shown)


gabehamilton (approved):

calebjacob (approved):
Thanks for the detailed context! Looks like a great solution. Nice job!

PR #726, opened 2024-03-28
feat: entity namespaces and stars

gabehamilton:
All entity framework data is now stored in the `social.near` contract under the `entities` key.
The next level below that is a namespace. Entities that don't specify a namespace will go under `default`. Next is the entity type then the entities themselves. So a  myCar `vehicle` entity in the `examples` namespace would be
```
<account_id>: { entities: { examples: { vehicle: { myCar: {...cardata}  } } } }
```
Staring an entity will use a `path` matching the same format.

### DIFF
--- a/indexers/agents/agents.js
+++ b/indexers/agents/agents.js
@@ -1,3 +1,4 @@
+// deprecated, delete once agiguild nexus is live
 import { Block } from "@near-lake/primitives";
 /**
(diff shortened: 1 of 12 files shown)


calebjacob (approved):
Looks good! Glad you were able to make use of `SocialIndexActionButton` :)

PR #731, opened 2024-03-29
feat: entity sort and search

gabehamilton:


calebjacob (approved):
Looks great! Thanks for using the `gateway-page-container` class :)

I'm not sure if it would make your life easier or not, but `DIG.InputSearch` has a debounce functionality built into it. Maybe give it a try and see if it's helpful!

gabehamilton:
> Looks great! Thanks for using the `gateway-page-container` class :)
> 
> I'm not sure if it would make your life easier or not, but `DIG.InputSearch` has a debounce functionality built into it. Maybe give it a try and see if it's helpful!

Nice. I'll try that out.

PR #734, opened 2024-04-02
feat: write only latest entity change, misc

gabehamilton:
* reduce user storage costs by only storing latest entity action
 * fix typo in components indexer
 * small updates to examples
 * Fetch entity details from QueryApi


jackson-harris-iii (approved):
LGTM!

PR #738, opened 2024-04-03
fix: dataplatform mispelling

Kevin101Zhang:


Kevin101Zhang:
Doing a larger review a bit be the end of this week and will restart the components indexer

gabehamilton:
Thanks. I just merged a PR that had the same change in it.

PR #739, opened 2024-04-03
fix: add overflow to username on notifications

shelegdmitriy:
I noticed this issue with very long username, so this PR introduces changes to fix this

Знімок екрана 2024-04-03 о 20 25 12


calebjacob (approved):
Nice fix!

PR #756, opened 2024-04-12
fix: notification counter

shelegdmitriy:
This PR introduces fix for new notifications count. Also, I've added a query to fetch for new notifications from queryApi and not from chain. Fallback logic also included

### DIFF
--- a/src/NearOrg/Notifications/NotificationButton.jsx
+++ b/src/NearOrg/Notifications/NotificationButton.jsx
@@ -59,18 +59,10 @@ const Button = ({ count }) => {
 };
 
 useEffect(() => {
-  const checkLastBlockHeight = setInterval(() => {
+  const checkNotificationsCountPeriod = setInterval(() => {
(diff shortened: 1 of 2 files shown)


calebjacob (approved):

src/NearOrg/Notifications/NotificationButton.jsx
```diff
@@ -58,36 +60,49 @@ const Button = ({ count }) => {
...
+        count
+      }
+    }
+  }
+`;
+
 useEffect(() => {
-  const checkLastBlockHeight = setInterval(() => {
+  const checkNotificationsCountPeriod = setInterval(() => {
+    const shouldFallback = Storage.get("notifications:shouldFallback", notificationFeedSrc);
```
gabehamilton, line 75: Is the 2nd arg a default?

calebjacob: It's actually the key of a different widget where the data is stored: https://thewiki.near.page/near.social_vm

PR #759, opened 2024-04-12
fix: entities indexer handles empty namespace key

gabehamilton:
An empty key in the data `"": ''` was causing the error.

Note: Working toward tests for the indexers.
Note: Tried to make this a clean training case. Issue->PR with test.
Note: code has been deployed to production through QueryApi

### DIFF
--- a/indexers/entities/entities_tests.js
+++ b/indexers/entities/entities_tests.js
@@ -1,0 +1,10 @@
+import getBlock from './entities.js';
+import testBlock from './test_block.js';
+
+describe('entities.js indexer', () => {
+
+    it('accepts data with an empty namespace', () => {
+        getBlock(testBlock);
+        // successful if it does not throw an error
+    });
+});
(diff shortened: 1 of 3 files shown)


indexers/entities/entities_tests.js
```diff
@@ -0,0 +1,10 @@
+import getBlock from './entities.js';
+import testBlock from './test_block.js';
+
+describe('entities.js indexer', () => {
+
+    it('accepts data with an empty namespace', () => {
+        getBlock(testBlock);
+        // successful if it does not throw an error
+    });
+});
```
shelegdmitriy, line 10: How to run this test locally?

gabehamilton: There's no harness for running it yet. I know that's odd. I included the test to move toward having tests and so that the PR makes a complete training case.

shelegdmitriy: Ok, got it!

shelegdmitriy (approved):

PR #760, opened 2024-04-15
Contact Us Page (1 of 2)

calebjacob:
Partially implements: https://github.com/near/near-discovery/issues/1064

Implements new Contact Us page for the gateway. We should merge this before the 2nd PR.

Part 2: https://github.com/near/near-discovery/pull/1115

### DIFF
--- a/src/Gateway/ContactUsPage.jsx
+++ b/src/Gateway/ContactUsPage.jsx
@@ -205,6 +205,8 @@ const faqs = [
 const Wrapper = styled.div`
   display: flex;
   flex-direction: column;
+  padding-left: 1.5rem;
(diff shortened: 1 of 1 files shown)


shelegdmitriy (approved):
LGMT! Just one question: are we good to use `<a>` tag instead of `<Link>` here?

calebjacob:
@shelegdmitriy It's a good question. I'm using simple `<a>` tags here since all of the links are external with `target="_blank"`.

shelegdmitriy:
> @shelegdmitriy It's a good question. I'm using simple `<a>` tags here since all of the links are external with `target="_blank"`.

Oh, makes sense!

PR #764, opened 2024-04-16
chore: add new items to Data Availability page

shelegdmitriy:
This PR introduces a couple of new items in Teams section:

Знімок екрана 2024-04-17 о 01 22 56

Closes: https://github.com/near/near-discovery/issues/1104

### DIFF
--- a/src/NearOrg/DataAvailabilityPage.jsx
+++ b/src/NearOrg/DataAvailabilityPage.jsx
@@ -16,6 +16,11 @@ const ipfsImages = {
     optimism: "bafkreidwitx5hu6hivyn3exi34moyea7livf6zfqk2dcny3z62ive7fpou",
     zeeve: "bafkreigwtxjuraug4l2rltpoy4wau6gaom6e52fwgyqfpdsh2t6voe6ypm",
     web3Game: "bafkreidbmy27gngqsz5opzy5dbhd7jdrnzt5r2e5grcpiad77zgjc4e6ki",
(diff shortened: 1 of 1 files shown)


src/NearOrg/DataAvailabilityPage.jsx
```diff
@@ -230,6 +235,12 @@ const LogoLinks = styled.div`
   }
 `;
 
+const LinkLogo = styled.a`
+  img {
+    filter: ${(p) => p.$grayscale ?? "none"};
```
shelegdmitriy, line 240: _"Why do we need this filter prop here?"_

Well, the Particle Network logo itself is very colorful. This hack + change at line 327 makes it dark grey easily

charleslavon (approved):

shelegdmitriy:
> https://github.com/near/nearorg_container

Thanks @calebjacob for mention this because I didn't know that. I can add the same to that repo. So, in the future, if we need to add something to one of marketing page we have to do this in two places, correct?

charleslavon:
Thanks for the reminder @calebjacob.  @shelegdmitriy that's correct Dima, so we should be more diligent to try and delay making updates such as this one for the next few weeks, so we can avoid duplicate work.

shelegdmitriy:
> Thanks for the reminder @calebjacob. @shelegdmitriy that's correct Dima, so we should be more diligent to try and delay making updates such as this one for the next few weeks, so we can avoid duplicate work.

@charleslavon just FYI: this page has been added and merged into `develop` branch of nearorg_container repo (so it's not live there but at least we won't forget about it 🙂)

PR #766, opened 2024-04-16
Entity tags

gabehamilton:


src/Entities/Template/Forms/FormBuilder.jsx
```diff
@@ -0,0 +1,310 @@
...
+  };
+
+  const onSubmitClick = () => {
+    if (onSubmit && isFormValid()) {
+      onSubmit(formFormattedValues);
+    }
+  };
+
+  return (
+    <div className="flex-grow-1 d-flex flex-column gap-4">
```
calebjacob, line 271: Just a heads up @gabehamilton, we're going to be moving away from Bootstrap with the new React on Chain engine. So let's try to stay away from using global Bootstrap classes like `d-flex`, etc.

calebjacob (approved):
Looks good! Left one comment. Feel free to address in a later PR.

src/Entities/Template/Forms/FormBuilder.jsx
gabehamilton: Cool. This component comes to us by way of DevHub. I was thinking it needs a rewrite anyway so I'll put in an Issue to move it off of bootstrap.

PR #785, opened 2024-04-25
feat: the Indexer status page uses new QueryApi status tables

gabehamilton:


src/Entities/QueryApi/DataPlatformIndexerStatuses.jsx
```diff
@@ -3,31 +3,68 @@ const GRAPHQL_ENDPOINT = props.GRAPHQL_ENDPOINT || "https://near-queryapi.api.pa
...
     }),
   });
 }
-const query = `query MyQuery {
-  indexer_state(where: {function_name: {_like: "${fullFilter}%"}}) {
-    status
-    current_block_height
-    function_name
+const query = (indexer) => `query MyQuery {
+  ${indexerAccountLink}_${indexer}_sys_metadata {
```
darunrs, line 61: I know none of the current dataplatform indexers have this problem, but the graphQL query may not work if it has symbols like hyphens, which are allowed in the indexer name but not the schema name. I'd run the returned names through some sanitization before putting them in the query. Here's how we do it currently:
```
const sanitizedAccountId = indexerDetails.accountId.replace(/[^a-zA-Z0-9]/g, '_').replace(/^([0-9])/, '_$1');
  const sanitizedIndexerName = indexerDetails.indexerName.replace(/[^a-zA-Z0-9]/g, '_').replace(/^([0-9])/, '_$1');
  const functionName = `${indexerDetails.accountId}/${indexerDetails.indexerName}`;
  const schemaName = `${sanitizedAccountId}_${sanitizedIndexerName}`;
  const tableName = `${schemaName}_sys_logs`;
```

src/Entities/QueryApi/DataPlatformIndexerStatuses.jsx
```diff
@@ -36,9 +73,12 @@ function handleResults(result) {
       }
       let data = result.body.data;
       if (data) {
-        const statuses = data.indexer_state;
+        const statuses = data[`${indexerAccountLink}_${indexer}_sys_metadata`];
         if (statuses.length > 0) {
-          setIndexerStatuses(statuses);
+          const status = statuses.find((s) => s.attribute === "STATUS")?.value;
+          const blockHeight = statuses.find((s) => s.attribute === "LAST_PROCESSED_BLOCK_HEIGHT")?.value;
```
darunrs, line 79: Should there be some fallback value for these? It's "possible" for these values to not be there, either because its provisioning, or if something went wrong in provisioning, or if the indexer hasn't run once yet.

src/Entities/QueryApi/DataPlatformIndexerStatuses.jsx
```diff
@@ -47,10 +87,27 @@ function handleResults(result) {
...
-fetchGraphQL(query).then((result) => handleResults(result));
+const update = () => {
+  asyncFetch("https://rpc.mainnet.near.org", {
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json",
+    },
+    body: JSON.stringify({
+      jsonrpc: "2.0",
+      id: "dontcare",
```
darunrs, line 98: LOL

darunrs (approved):
Super cool work you did so quickly! Some comments but overall looks good!

src/Entities/QueryApi/DataPlatformIndexerStatuses.jsx
gabehamilton: I think they'll just be missing in the UI if they are missing in the data.

src/Entities/QueryApi/DataPlatformIndexerStatuses.jsx
gabehamilton: that's the standard in all our docs :-)

PR #792, opened 2024-04-29
Daily release

gabehamilton:


src/Entities/Template/EntityDetails.jsx
```diff
@@ -132,14 +131,50 @@ const entityProperties = (obj) => {
...
+            case "image":
+              return (
+                <>
+                  <Text bold key={key}>
+                    {capitalize(key)}:
+                  </Text>
+                  <PropValue>
+                    <img className="logo" src={imageUrl} alt={key} />
+                  </PropValue>
+                </>
```
shelegdmitriy, line 148: The `key` property should be added at the top level. This means this section should be look like this:
```suggestion
                <Fragment key={key}>
                  <Text bold>
                    {capitalize(key)}:
                  </Text>
                  <PropValue>
                    <img className="logo" src={imageUrl} alt={key} />
                  </PropValue>
                </Fragment>
```
`<Fragment>` and `<>` are both fragments. I'm fine with moving on with this code, but it would be good to have this change as a separate PR

shelegdmitriy (approved):

src/Entities/Template/EntityDetails.jsx
gabehamilton: Nice catch, thanks! I'll do it in a separate PR to keep main & develop in sync.

PR #805, opened 2024-05-03
fix: clean up 

gabehamilton:
* clean up copy/share urls
 * summary component accepts action param (eliminates custom summary components)
 *  ipfs image support in EntitySummary

### DIFF
--- a/src/Entities/Template/EntityDetails.jsx
+++ b/src/Entities/Template/EntityDetails.jsx
@@ -71,7 +71,7 @@ if (error) {
 if (!entity) {
   return <p>Loading...</p>;
 }
-const { title } = schema;
+const { entityTitle } = schema;
 
 const editType = accountId === context.accountId ? "edit" : "fork";
 const editLabel = editType === "edit" ? "Edit" : "Fork";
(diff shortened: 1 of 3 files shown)


calebjacob (approved):

src/Entities/Template/EntityDetails.jsx
```diff
@@ -186,6 +186,11 @@ const entityProperties = (obj) => {
 const listLink = href({
   widgetSrc: returnTo,
 });
+const detailsLink = href({
+  widgetSrc: `${REPL_ACCOUNT}/widget/Entities.Template.EntityDetails`,
+  params: { src, schemaFile, namespace, entityType, returnTo: homeLink },
+});
+const detailsUrl = "https://${REPL_NEAR_URL}/" + detailsLink;
```
shelegdmitriy, line 193:
```suggestion
const detailsUrl = `https://${REPL_NEAR_URL}/${detailsLink}`;
```

This suggestion will work the same but looks cleaner

src/Entities/Template/EntitySummary.jsx
```diff
@@ -258,6 +261,28 @@ return (
             ),
           }}
         />{" "}
+        {entityActionUrlFunction && (
+          <Widget
+            src="${REPL_ACCOUNT}/widget/DIG.Tooltip"
+            props={{
+              content: "Use " + title,
```
shelegdmitriy, line 268:
```suggestion
              content: `Use ${title}`,
```

shelegdmitriy (approved):
LGTM! Added a few comments

src/Entities/Template/EntityDetails.jsx
gabehamilton: I used concatenation here to highlight that the replacement of REPL_NEAR_URL is not via ES6 template. Also, it should be slightly faster, although we don't really care about performance here.

PR #825, opened 2024-06-04
fix: search input focus issue

shelegdmitriy:
This PR addresses an issue where the search bar frequently loses focus. This problem was causing frustration for users who needed to repeatedly click on the input bar to continue typing and see auto-suggestions.

Check out this [message](https://pagodaplatform.slack.com/archives/C0559BC3WBZ/p1717521230244739) for additional context.


calebjacob (approved):
Nice fix!

PR #832, opened 2024-06-10
fix: Gateway Homepage

gagdiez:
The [homepage](https://dev.near.org) is currently failing to load the trending apps from nearcatalog. Apparently nearcatalog's API has changed and returns the apps within a `data` field when requesting the most used apps:

https://nearcatalog.xyz/wp-json/nearcatalog/v1/projects-by-category?cid=trending


shelegdmitriy (approved):
Thanks for the catch!

PR #836, opened 2024-06-25
chore: reduce amount of search categories (Part 1/2)

shelegdmitriy:
As per discussion with @charleslavon we decided to proceed with this https://github.com/near/near-discovery/issues/1196 issue by this way:

1. Remove search categories such as "posts", "posts-and-comments", "people" and leave **only** components including nearcatalog and widgets (about widgets I'm not sure, because don't remember);
2. Bring the top navbar back along with search (this is the **part 2** on the near-discovery side);

Closes https://github.com/near/near-discovery/issues/1196

Feel free to leave any questions you have!

### DIFF
--- a/src/Navigation/NameDropdown.jsx
+++ b/src/Navigation/NameDropdown.jsx
@@ -49,7 +49,7 @@ return (
   <Widget
     src="${REPL_ACCOUNT}/widget/DIG.DropdownMenu"
     props={{
-      trigger: profile.name,
+      trigger: profile.name || accountId,
       items: menuItems,
     }}
   />
--- a/src/Navigation/NameDropdown.jsx
+++ b/src/Navigation/NameDropdown.jsx
@@ -46,13 +46,11 @@ const menuItems = [
 ];
 
 return (
-    <Widget
-      src="${REPL_ACCOUNT}/widget/DIG.DropdownMenu"
-      props={{
-        trigger: (
-          profile.name
-        ),
-        items: menuItems,
-      }}
-    />
+  <Widget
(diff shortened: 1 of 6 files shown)


calebjacob (approved):

src/Search/IndexPage.jsx
```diff
@@ -3,7 +3,7 @@ const APPLICATION_ID = props.appId ?? "B6PI9UKKJT";
 const INDEX = props.index ?? "replica_prod_near-social-feed";
 const API_URL = props.apiUrl ?? `https://${APPLICATION_ID}-dsn.algolia.net/1/indexes/${INDEX}/query?`;
 const INITIAL_PAGE = props.initialPage ?? 0;
-const facets = props.facets ?? ["All", "People", "Apps", "Components", "Posts"];
+const facets = props.facets ?? ["All", "Apps", "Components"];
```
gagdiez, line 6: what is our stand on Components? Because we said in the `docs` that we are deprecating them

Do we want people to find random components from random users when we advise against it security on docs and the upcoming blogpost?

src/Navigation/NameDropdown.jsx
```diff
@@ -46,13 +46,11 @@ const menuItems = [
...
-        trigger: (
-          profile.name
-        ),
-        items: menuItems,
-      }}
-    />
+  <Widget
+    src="${REPL_ACCOUNT}/widget/DIG.DropdownMenu"
+    props={{
+      trigger: profile.name,
```
gagdiez, line 52:
```suggestion
      trigger: profile.name || accountId,
```

PR #838, opened 2024-06-26
fix: display accountId when profile don't exist

gagdiez:
Currently, if the user has not created a profile, the menu will display "Trigger" since that is the default text displayed by `DIG.DropdownMenu`

This PR fixes it by displaying the AccountId if no profile exist


gagdiez:
added this change to #836

PR #841, opened 2024-07-08
WIP: Add docs in search

gagdiez:
This PR adds the ability to search for docs directly on the general searchbar

We might want to discuss:
- Removing the "All" tab
- Changing the "Apps" tab, so the result comes from NEAR Catalog instead of showing only components

### DIFF
--- a/src/Search/TypeAheadDropdown.jsx
+++ b/src/Search/TypeAheadDropdown.jsx
@@ -1,13 +1,43 @@
-const SEARCH_API_KEY = props.searchApiKey ?? "fc7644a5da5306311e8e9418c24fddc4";
-const APPLICATION_ID = props.appId ?? "B6PI9UKKJT";
-const INDEX = props.index ?? "replica_prod_near-social-feed";
-const API_URL = props.apiUrl ?? `https://${APPLICATION_ID}-dsn.algolia.net/1/indexes/${INDEX}/query?`;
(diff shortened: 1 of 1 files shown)


shelegdmitriy (changes requested):
Overall looks good! Thank you @matiasbenary for contribution! But I want to highlight that all this changes should be added to `src/Search/IndexPage.jsx` page also.

> Removing the "All" tab

Well, I think we can remove this tab from `TypeAheadDropdown` but let's leave it on `src/Search/IndexPage.jsx`. There is enough space for that there. WDYT?

> Changing the "Apps" tab, so the result comes from NEAR Catalog instead of showing only components

Apps tab already has the nearcatalog records if I'm not mistaken 🤔

Also, I do understand that this PR is in progress but I think this UI could be improved a bit:
Знімок екрана 2024-07-10 о 17 22 11

1.  Let's display a doc's category instead of `COMPONENTS 2389` (or Docs after changes). Seems like we do have such data so why don't we use it?
2. Let's add a hover effect like `text-decoration: underline` for title or something

shelegdmitriy:
Hey @matiasbenary and @gagdiez! Seems like there is no need in this PR in favour of this https://github.com/near/near-discovery/pull/1249 right?

gagdiez:
@shelegdmitriy indeed, closing this one

PR #846, opened 2024-07-29
Revert "chore: adds an invisible link to be used by playwright automation"

charleslavon:
Reverts near/near-discovery-components#845

I think I can accomplish this directly via playwright code ```await page.setContent(`<style>
      a {
        zoom: 100%
      }
      </style>
      <a href="/sandbox/>
      `);


charleslavon:
actually this is needed

